PR #72, opened 2022-09-06
Proposal: introduce option to always peek

JoviDeCroock:
This introduces an option to always peek so we don't subscribe, the use case for this would be distributive repetitive renderers like render-to-string and prepass.

This currently only affects `text` optimization but this should probably apply to all of these 😅 another alternative would be to register an option to `unsubscribeAll`


JoviDeCroock:
Superseded by https://github.com/preactjs/preact-render-to-string/pull/242

PR #73, opened 2022-09-06
fix `peak()` example in readme

tjenkinson:
I think you meant to set the `value`?


marvinhagemeister (approved):
Good catch, thanks for the PR 👍

Issue #74, opened 2022-09-06
`renderToString` does not properly serialize `<input checked={signal} />`


lucacasonato:
```
import { signal } from "@preact/signals";
import { renderToString } from "preact-render-to-string";

const s = signal(false);
const html = renderToString(<input checked={s} />);
console.log(html);
```

This prints `<input checked="false" />`.

If I do `s.value` it prints `<input />`.

PR #76, opened 2022-09-06
Correctly peek and replace value

JoviDeCroock:
Before we never unwrapped the signal passed into the element-attributes, this lead to the issue described in #74. We resolve this by assigning the prop to `signal.peek()`, this however leads to us losing the reference to the signal on `vnode.props[attribute-key]` which is used for the updater. To resolve the updater losing signal reference we switch the underlying data-structure to use an object of `{ key signal }` to save the reference.

While writing tests I noticed that the updater was never properly called as it would call the updater of the passed-in signal rather than the one we created with `createUpdater`, this is fixed by inheriting the `_updater` function to the props-signal.

Fixes https://github.com/preactjs/signals/issues/74
CSB Repro: https://codesandbox.io/s/jovial-nova-8xgk76?file=/src/signals.js:0-3518

### DIFF
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -65,6 +65,7 @@ function getElementUpdater(vnode: VNode) {
 			for (let i = 0; i < signalProps.length; i++) {
 				let { _key: prop, _signal: signal } = signalProps[i];
 				let value = signal._value;
+				if (!dom) return;
 				if (prop in dom) {
 					// @ts-ignore-next-line silly
 					dom[prop] = value;
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -64,7 +64,7 @@ function getElementUpdater(vnode: VNode) {
 
 			for (let i = 0; i < signalProps.length; i++) {
 				let { _key: prop, _signal: signal } = signalProps[i];
-				let value = signal.peek();
+				let value = signal._value;
 				if (prop in dom) {
 					// @ts-ignore-next-line silly
 					dom[prop] = value;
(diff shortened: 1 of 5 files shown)


packages/preact/src/index.ts
```diff
@@ -152,7 +151,8 @@ hook(OptionsTypes.DIFF, (old, vnode) => {
 				// first Signal prop triggers creation/cleanup of the updater:
 				if (!updater) updater = getElementUpdater(vnode);
 				// track which props are Signals for precise updates:
-				updater._props.push(i);
+				updater._props.push({ _key: i, _signal: value });
+				props[i] = value.peek()
```
marvinhagemeister, line 155: Just checked and this is the only change needed to make the new test case pass.

JoviDeCroock: Yes, need to add a test with a subsequent update

JoviDeCroock: @marvinhagemeister peculiarly the test still fails and we never enter the `updater` function 😅

marvinhagemeister (approved):

packages/preact/src/index.ts
```diff
@@ -152,7 +152,18 @@ hook(OptionsTypes.DIFF, (old, vnode) => {
...
 				// track which props are Signals for precise updates:
-				updater._props.push(i);
+				updater._props.push({ _key: i, _signal: value });
+				let newUpdater = updater._updater
+				if (value._updater) {
+					let oldUpdater = value._updater
+					value._updater = () => {
+						newUpdater();
+						oldUpdater();
+					}
```
developit, line 162: I think this might create infinite updaters. If we're setting `_props`, how come we need to run the old updater?

JoviDeCroock: Signals reused fot multi attributes test 3 I added

developit: Sorry, I meant when vnodes change. In X, `vnode` is unique every time this hook is called.

PR #82, opened 2022-09-07
chore: Splits README and ensures each pkg is published with one

rschristian:
None of the packages currently are published with ReadMes, which makes it difficult to browse on NPM.

This does 2 things:
 - Copies root ReadMe to core on `prepublish`
 - Splits the Preact & React content into their own ReadMes

I kept the intro blurb in each, not sure how others feel about that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,11 +25,11 @@ npm install @preact/signals-react
 	- [`computed(fn)`](#computedfn)
 	- [`effect(fn)`](#effectfn)
 	- [`batch(fn)`](#batchfn)
-- [Preact Integration](./packages/preact/README.md)
+- [Preact Integration](./packages/preact/README.md#preact-integration)
 	- [Hooks](./packages/preact/README.md#hooks)
 	- [Rendering optimizations](./packages/preact/README.md#rendering-optimizations)
 		- [Attribute optimization (experimental)](./packages/preact/README.md#attribute-optimization-experimental)
-- [React Integration](./packages/react/README.md)
+- [React Integration](./packages/react/README.md#react-integration)
 	- [Hooks](./packages/react/README.md#hooks)
 - [License](#license)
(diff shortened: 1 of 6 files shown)


packages/core/package.json
```diff
@@ -7,7 +7,10 @@
 	"authors": [
 		"The Preact Authors (https://github.com/preactjs/signals/contributors)"
 	],
-	"repository": "preactjs/signals",
+	"repository": {
+		"type": "git",
+		"url": "https://github.com/preactjs/signals"
+	},
```
rschristian, line 13: From what I understand, in order for NPM to work with relative links in our ReadMes, we need to update the specifier here.

[Source](https://stackoverflow.com/a/68738752/15388164)

README.md
```diff
@@ -25,12 +25,12 @@ npm install @preact/signals-react
...
-		- [Attribute optimization (experimental)](#attribute-optimization-experimental)
-- [React Integration](#react-integration)
-	- [Hooks](#hooks-1)
+- [Preact Integration](./packages/preact/README.md#preact-integration)
+	- [Hooks](./packages/preact/README.md#hooks)
+	- [Rendering optimizations](./packages/preact/README.md#rendering-optimizations)
+		- [Attribute optimization (experimental)](./packages/preact/README.md#attribute-optimization-experimental)
+- [React Integration](./packages/react/README.md#react-integration)
+	- [Hooks](./packages/react/README.md#hooks)
 - [License](#license)
```
rschristian, line 34: Is it weird if this links to other docs? It's sorta laid out as a table of contents, but if we link elsewhere, then it stops being the contents of this doc.

marvinhagemeister: I think it's fine

marvinhagemeister (approved):

PR #85, opened 2022-09-07
Add unwrapping signals

intrnl:
Adds an utility function to unwrap signals, useful if you want to accept both signals and non-signals, and only want the actual value.

Peeking isn't used here because the intention is that you'd still want to have your signal tracked in the case that it is one.

### DIFF
--- a/.changeset/hungry-dancers-eat.md
+++ b/.changeset/hungry-dancers-eat.md
@@ -1,0 +1,7 @@
+---
+"@preact/signals-core": minor
+"@preact/signals": minor
+"@preact/signals-react": minor
+---
+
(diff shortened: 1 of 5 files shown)


marvinhagemeister:
Thanks for making a PR! Great to see you around here 👍 

Our goal with Signals is to keep the API as small as possible only add functions that are deemed essential. An `unwrap()` function is useful - no doubt - but I believe that it's not something every project needs. If the need arises in a project, then I feel like they are well served with adding that little utility function in their project.

intrnl:
Gotcha, I think I'll close the PR for the time being :+1:

Issue #89, opened 2022-09-07
Typo in invalidated batch example in README


kim3er:
In the following example:

```
import { signal, computed, effect, batch } from "@preact/signals-core";

const counter = signal(0);
const double = computed(() => counter.value * 2);
const tripple = computed(() => counter.value * 3);

effect(() => console.log(double.value, tripple.value));

batch(() => {
	counter.value = 1;
	// Logs: 2, despite being inside batch, but `tripple`
	// will only update once the callback is complete
	console.log(counter.double);
});
// Now we reached the end of the batch and call the effect
```

I think `counter.double` should read `double.value`.

marvinhagemeister:
Fixed by #90

PR #90, opened 2022-09-07
Fixed typo in README

kim3er:
Relating to https://github.com/preactjs/signals/issues/89.

Changed `counter.double` to read `double.value`.


marvinhagemeister (approved):
Good catch! Thank you so much for the PR 🙌

PR #91, opened 2022-09-07
Proposal: use signal effect

JoviDeCroock:
Currently running effects is a bit of a pitfall imho, or we run them globally outside of the function-scope, or we have to wrap them with a `useMemo`/`useEffect`/... This lead me on the path of creating a small utility function that abstracts this. When people want to use a global `signal`/`effect` they can import from the `core` package, however when they want a component-specific signal they leverage `useSignal`. It's here where I saw a discrepancy, we have a utility to help provide a signal but no utility for the effect.

Here comes the introduction of `useSignalEffect` it basically wraps an `effect` with `useEffect` and holds a stable ref to the callback being passed to the `useSignalEffect` so we don't have to bother adding dependencies to the array of the used `useEffect`. Currently the only short-coming I may be foreseeing is the lack of an `onInvalidate` function, i.e. what happens if an effect updates that is for example still fetching, we could be introducing a race-condition there.

Usage:

```js
const Results = () => {
  const page = useSignal(1);
  
  useSignalEffect(() => {
    fetch('x?page=' + page.value).then(updateResults)
  })
  
  return results
}
```

### DIFF
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -323,13 +323,20 @@ export function useComputed<T>(compute: () => T) {
 	return useMemo(() => computed<T>(() => $compute.current()), []);
 }
 
-export function useSignalEffect(cb: () => void) {
+export function useSignalEffect(cb: () => void | (() => void)) {
 	const callback = useRef(cb);
 	callback.current = cb;
 
-	useEffect(() => effect(() => {
-		callback.current();
-	}), []);
+	useEffect(() => {
+		let cleanup: (() => void) | undefined;
+		effect(() => {
+			if (cleanup) {
+				cleanup();
+				cleanup = undefined
+			}
+			cleanup = callback.current();
+		})
+	}, []);
 }
 
 /**
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -329,7 +329,7 @@ export function useSignalEffect(cb: () => void | (() => void)) {
 
 	useEffect(() => {
 		let cleanup: (() => void) | undefined;
-		return effect(() => {
+		effect(() => {
 			if (cleanup) {
 				cleanup();
 				cleanup = undefined
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -328,7 +328,7 @@ export function useSignalEffect(cb: () => void) {
 	callback.current = cb;
 
 	useEffect(() => {
-		effect(() => {
+		return effect(() => {
 			callback.current();
 		})
 	}, [])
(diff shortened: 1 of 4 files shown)


packages/preact/src/index.ts
```diff
@@ -1,5 +1,5 @@
 import { options, Component, createElement } from "preact";
-import { useRef, useMemo } from "preact/hooks";
+import { useRef, useMemo, useEffect } from "preact/hooks";
```
developit, line 2:
```suggestion
import { useRef, useMemo, useLayoutEffect } from "preact/hooks";
```

packages/preact/src/index.ts
```diff
@@ -283,6 +283,17 @@ export function useComputed<T>(compute: () => T) {
...
+export function useSignalEffect(cb: () => void) {
+	const callback = useRef(cb);
+	callback.current = cb;
+
+	useEffect(() => {
+		return effect(() => {
+			callback.current();
+		})
+	}, [])
+}
```
developit, line 295:
```suggestion
export function useSignalEffect(cb: () => void) {
	const callback = useRef(cb);
	callback.current = cb;

	useLayoutEffect(() => effect(() => {
		callback.current();
	}), []);
}
```

packages/preact/src/index.ts
```diff
@@ -283,6 +283,15 @@ export function useComputed<T>(compute: () => T) {
 	return useMemo(() => computed<T>(() => $compute.current()), []);
 }
 
+export function useSignalEffect(cb: () => void) {
```
marvinhagemeister, line 286: Let's add this hook to the react adapter too 👍

marvinhagemeister (approved):

packages/preact/src/index.ts
```diff
@@ -323,6 +323,25 @@ export function useComputed<T>(compute: () => T) {
 	return useMemo(() => computed<T>(() => $compute.current()), []);
 }
 
+export function useSignalEffect(cb: () => void | (() => void)) {
```
motss, line 326: @JoviDeCroock is the param type duplicated?

packages/react/src/index.ts
```diff
@@ -162,3 +163,22 @@ export function useComputed<T>(compute: () => T) {
 	$compute.current = compute;
 	return useMemo(() => computed<T>(() => $compute.current()), []);
 }
+
+export function useSignalEffect(cb: () => void | (() => void)) {
```
motss, line 167: @JoviDeCroock Param type looks duplicated

packages/preact/src/index.ts
JoviDeCroock: no you can return void or a function that returns void i.e. a clenaup function

PR #96, opened 2022-09-07
Fix swapping and HMR for signals when used as text

developit:
The HMR problem happens because `<Text data={signal} />` hadn't accounted for the `data` prop being given an entirely new Signal reference. This PR adds that logic.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,10 +1,5 @@
 lockfileVersion: 5.4
 
-patchedDependencies:
-  microbundle@0.15.1:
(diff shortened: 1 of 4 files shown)


developit:
CI error is a bug in compressed-size-action with pnpm.

marvinhagemeister (approved):

pnpm-lock.yaml
```diff
@@ -1,10 +1,5 @@
 lockfileVersion: 5.4
 
-patchedDependencies:
```
JoviDeCroock, line 3: Is this intended

PR #117, opened 2022-09-10
Remove root signal

developit:
This fixes a memory leak caused by all signals being addressable from our internal `ROOT` signal. We never ended up using the root signal, and removing it means computed graphs within components get nicely GC'd when unmounted.


JoviDeCroock (approved):

marvinhagemeister (approved):
Nice 👍

PR #121, opened 2022-09-11
Base `.value` on `.peek()`

marvinhagemeister:
The only difference between `.value` and `.peek()` is that `.value` should set up a subscription, whereas `.peek()` shouldn't. This means `.value` can be based on `.peek()`. Both were looking already pretty similar in their implementation.

### DIFF
--- a/.changeset/silent-schools-begin.md
+++ b/.changeset/silent-schools-begin.md
@@ -1,0 +1,5 @@
+---
+"@preact/signals-core": patch
+---
+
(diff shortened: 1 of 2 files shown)


developit:
Since this is slightly larger, we should base the decision on performance. My fear is that we are doubling the function calls for every .value access.

marvinhagemeister:
The implementation in `main` is shortened a bit so this PR is redundant.

PR #123, opened 2022-09-11
Small fixes to core

elliotwaite:
Just some suggestions I noticed while reading over the code. Mostly typo fixes in comments, but I also thought using the parameter name `dep` instead of `sub` in those arrow functions made more sense, and then I noticed that in one of the tests, the code didn't seem to align with the comment at the top of the test, it looks like that `eSpy` function should be reading `d.value`.

Thanks for building this library. I'm looking forward to trying it out some more.

### DIFF
--- a/.changeset/selfish-cameras-play.md
+++ b/.changeset/selfish-cameras-play.md
@@ -1,0 +1,5 @@
+---
+"@preact/signals-core": patch
+---
+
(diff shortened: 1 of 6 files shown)


marvinhagemeister (approved):
Awesome, thanks for making a PR! A quick rebase and it's ready to go in 🎉  Feel free to add a changeset via `pnpm changest` which will be used for the release notes 👍

Issue #115, opened 2022-09-09
signals in react (three-fiber) doesn't work transiently


drcmda:
codesandbox: https://codesandbox.io/s/interesting-panna-urdtu6?file=/src/App.js

it's not a div, but it would be fantastic if it could work.

CodyJasonBennett:
This works fine when using `React.createElement` in development ([there's current a hard-dependency on that to change `opacity` to `opacity.value` at runtime](https://github.com/preactjs/signals/blob/d3c8cf91f96bb2d186d9e2524c9c67fcf77fee93/packages/react/src/index.ts#L25)):

```jsx
import * as React from 'react'
import * as ReactDOM from 'react-dom/client'
import { Canvas } from '@react-three/fiber'
import { signal } from '@preact/signals-react'

const opacity = signal(1)
setTimeout(() => (opacity.value = 0.5), 3000)

function Box() {
  console.log('rerender')
  return React.createElement(
    'mesh',
    null,
    React.createElement('boxGeometry'),
    React.createElement('meshBasicMaterial', { transparent: true, color: 'orange', opacity }),
  )
}

ReactDOM.createRoot(window.root).render(React.createElement(Canvas, null, React.createElement(Box)))

```

I found some other issues that completely break the bindings with a fix in #130, but I don't believe they are directly related.

developit:
The reason this doesn't work is because we don't currently support fine-grained prop updates in the React integration. The Preact integration supports this because it injects fine-grained prop-to-DOM updates, but we cannot do that in React.

However, it is possible to implement VDOM-based fine-grained updates in a way that works nicely with `react-three-fiber` as well as React-DOM. Here is a patch that makes this work: (just import anywhere)

```js
import { Signal } from '@preact/signals-react'
import * as rt from 'react/jsx-runtime'
import * as React from 'react'
rt.jsx = wrap(rt.jsx)
rt.jsxs = wrap(rt.jsxs)
let createElement = React.createElement
React.createElement = wrap(createElement)
const wrappers = new Map()
function wrap(method) {
  return function (type) {
    if (typeof type === 'string') {
      let t = wrappers.get(type)
      if (!t) wrappers.set(type, (t = Wrapper.bind(null, type)))
      arguments[0] = t
    }
    return method.apply(this, arguments)
  }
}
function Wrapper(type, props) {
  let p = {}
  for (let i in props) {
    let v = props[i]
    p[i] = i !== 'children' && v instanceof Signal ? v.value : v
  }
[...]

PR #130, opened 2022-09-12
fix(react): track owners separately, mutate updaters with dispatcher

### LINKED ISSUE #70: React.StrictMode misses state update (using with ViteJS)
I know this is super early days and this looks really exciting!

I tried setting up a new TypeScript + React project from scratch with ViteJS and using your counter example the first count+ click doesn't update the UI. However the second does so counter goes from 0->2->3 etc. When disabling React.StrictMode it works however. Just a heads up. Thank you for this project!

FYI: Uses React 18.2.

### LINKED ISSUE #125: @preact/signals-react crashes on production build, at least using `react-scripts build`
cf: https://github.com/preactjs/signals/issues/86

The error here is:  `TypeError: Invalid value used as weak map key`

image

What appears to be the case is that `internals.ReactCurrentOwner.current` is not set.

Here are the versions I have installed:
```
    "@preact/signals-react": "^1.0.0",
    "react": "^18.2.0",
    "react-dom": "^18.2.0",
    "react-scripts": "5.0.1",
    "typescript": "^4.4.2",
```
and the browser configuration:
```
  "browserslist": {
    "production": [
      ">0.2%",
      "not dead",
      "not op_mini all"
    ],
    "development": [
      "last 1 chrome version",
      "last 1 firefox version",
      "last 1 safari version"
    ]
  }
```

### How to reproduce:

1. using `create-react-app` (`typescript` template), create a react app
2. use `@preact/signals-react`
3. `yarn build` (creates a production build)
4. open your built browser-app in the browser
5. behold this error

### LINKED ISSUE #135: React skips the first rerender in StrictMode
react: `18.2.0`
@preact/signals-react: `1.0.0`

When `<StrictMode />` is wrapping `<App />`, the first rerender of every signal-attached component is ignored.

To reproduce:

1. Enter https://codesandbox.io/s/amazing-glitter-6eyndo?file=/src/index.js
2. Click on one of the buttons.
3. You can see that the new value logs correctly, but React simply won't repaint the screen.
4. Click on the same button again and react will "skip" and render the new value (the previous one was ignored).
5. Now remove StrictMode wrapping from `index.js` and repeat the previous steps, everything should work properly.

I personally don't use StrictMode, I just wanted to let you know that this problem exists.

Thanks for this amazing project, it really is revolutionary!

CodyJasonBennett:
Fixes #125
Fixes #135
Fixes #70

This PR tracks `ReactCurrentOwner` separately from `ReactCurrentDispatcher` since the former is not guaranteed to be populated when `ReactCurrentDispatcher` is mutated. This is very evident from #125, and you can verify by running against a production build of React. Annoyingly, this behavior is different than that of a development build of React.

Additionally, signals' updaters are mutated on subsequent runs since there are multiple dispatchers during the lifecycle of a component, and holding on to a stale dispatcher can break the re-rendering mechanism (it wouldn't do anything).

> **Note**: The relevant dispatcher states are [invalid](https://github.com/facebook/react/blob/c156ecd483b4adc8e16254b8cab3ee735e893271/packages/react-reconciler/src/ReactFiberHooks.old.js#L2434), [mount](https://github.com/facebook/react/blob/c156ecd483b4adc8e16254b8cab3ee735e893271/packages/react-reconciler/src/ReactFiberHooks.old.js#L2468), [update](https://github.com/facebook/react/blob/c156ecd483b4adc8e16254b8cab3ee735e893271/packages/react-reconciler/src/ReactFiberHooks.old.js#L2501), [re-render](https://github.com/facebook/react/blob/main/packages/react-reconciler/src/ReactFiberHooks.old.js#L2535) -- invalid and re-render are skipped by `isInvalidHookAccessor` and `lock`, respectively. These are unique per reconciler, so #115 would have two pairs of mount and update dispatchers.

The changes in this PR originally came from tweaking with v1 JS output, which you can find [in this gist](https://gist.github.com/CodyJasonBennett/3fb8c092db76da4761e53908752340b1). This is runnable if you want to copy/paste it into a Codesandbox, although ensure you're using the old JSX transform (`React.createElement`) if possible for the moment.

### DIFF
--- a/packages/react/src/index.ts
+++ b/packages/react/src/index.ts
@@ -126,6 +126,7 @@ Object.defineProperty(internals.ReactCurrentDispatcher, "current", {
 			if (!updater || !lastDispatcher !== api) {
 				updater = createUpdater(rerender);
 				updaterForComponent.set(lastOwner, updater);
+				lastDispatcher = api;
 			}
 			setCurrentUpdater(updater);
 		} else {
--- a/packages/react/src/index.ts
+++ b/packages/react/src/index.ts
@@ -89,10 +89,8 @@ Object.defineProperties(Signal.prototype, {
 	ref: { value: null },
 });
 
-// Stores all tracked component Fibers (note: this is never cleared and will leak memory)
-const componentQueue = new Set<ReactOwner>();
-
 // Track the current owner (roughly equiv to current vnode)
+let lastComponent: ReactOwner;
 let currentOwner: ReactOwner;
 Object.defineProperty(internals.ReactCurrentOwner, "current", {
 	get() {
@@ -100,20 +98,17 @@ Object.defineProperty(internals.ReactCurrentOwner, "current", {
 	},
 	set(owner) {
 		currentOwner = owner;
-		if (currentOwner && !componentQueue.has(currentOwner))
-			componentQueue.add(currentOwner);
+		if (currentOwner) lastComponent = currentOwner;
 	},
 });
 
 // Tracks component updaters per dispatcher
-const dispatcherQueue = new WeakMap<
-	ReactDispatcher,
-	WeakMap<ReactOwner, Updater>
->();
+const updaterForComponent = new WeakMap<ReactOwner, Updater>();
 
 // Track the current dispatcher (roughly equiv to current component impl)
 let lock = false;
 const UPDATE = () => ({});
+let lastDispatcher: ReactDispatcher;
 let currentDispatcher: ReactDispatcher;
 Object.defineProperty(internals.ReactCurrentDispatcher, "current", {
 	get() {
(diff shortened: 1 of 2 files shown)


packages/react/src/index.ts
```diff
@@ -91,15 +91,22 @@ Object.defineProperties(Signal.prototype, {
...
+let currentOwner: ReactOwner;
+Object.defineProperty(internals.ReactCurrentOwner, "current", {
+	get() {
+		return currentOwner;
+	},
+	set(owner) {
+		currentOwner = owner;
+		if (currentOwner) lastOwner = currentOwner;
+	},
+});
```
CodyJasonBennett, line 104: `currentOwner` would be better defined as `ReactOwner | null` as per the `React.RefObject` interface.

packages/react/src/index.ts
```diff
@@ -108,17 +115,18 @@ Object.defineProperty(internals.ReactCurrentDispatcher, "current", {
...
-			const currentOwner = internals.ReactCurrentOwner.current;
-			let updater = updaterForComponent.get(currentOwner);
-			if (!updater) {
+
+			let updater = updaterForComponent.get(lastOwner);
+			if (!updater || lastDispatcher !== api) {
 				updater = createUpdater(rerender);
-				updaterForComponent.set(currentOwner, updater);
+				updaterForComponent.set(lastOwner, updater);
+				lastDispatcher = api;
```
CodyJasonBennett, line 129: I haven't looked too much into the core signals mechanism but if signals were only cached after the first run and `signal._updater` was mutated in subsequent runs, that seems to play nice with #70 or #135.

```js
let updater = updaterForComponent.get(lastOwner);
if (!updater) {
  updater = createUpdater(rerender);
  updaterForComponent.set(lastOwner, updater);
} else {
  updater._updater = rerender;
}
```

marvinhagemeister (commented):
This looks great! Let's add one or few test cases that ensure that we don't regress on this in the future 👍

CodyJasonBennett:
Would it be fine to settle for cases that cover the linked issues (doesn't crash in production, will consistently re-render in both production/strict-mode) or would you prefer to re-run the test suite (IDK how you'd sanely configure a testing matrix)?

marvinhagemeister:
> Would it be fine to settle for cases that cover the linked issues

Sounds good enough to me 👍

marvinhagemeister (approved):
This is great! Thank you for your PR 🙌

CodyJasonBennett:
We'll have to trust Netlify CI in lieu of a controlled production environment.

PR #129, opened 2022-09-12
Small fixes to core

elliotwaite:
Just some suggestions I noticed while reading over the code. Mostly typo fixes in comments, but I also thought using the parameter name dep instead of sub in those arrow functions made more sense, and then I noticed that in one of the tests, the code didn't seem to align with the comment at the top of the test, it looks like that eSpy function should be reading d.value.

Thanks for building this library. I'm looking forward to trying it out some more.

Note: This PR is a redo of https://github.com/preactjs/signals/pull/123

PR #131, opened 2022-09-12
Make activation and signal updates one and the same

### LINKED ISSUE #124: A computed value is updated even when all of its dependencies haven't changed.
From reading over the logic in [core](https://github.com/preactjs/signals/blob/main/packages/core/src/index.ts), I think I found something that could go wrong. I created this test, which currently isn't passing, to illustrate the issue:
```
it("should not update a sub if all deps unmark it", () => {
    // In this scenario "B" and "C" always return the same value. When "A"
    // changes, "D" should not update.
    //     A
    //   /   \
    // *B     *C
    //   \   /
    //     D
    const a = signal("a");
    const b = computed(() => {
        a.value;
        return "b";
    });
    const c = computed(() => {
        a.value;
        return "c";
    });
    const spy = sinon.spy(() => b.value + " " + c.value);
    const d = computed(spy);
    expect(d.value).to.equal("b c");
    spy.resetHistory();

    a.value = "aa";
    expect(spy).not.to.be.called; // <- this assertion currently does not pass.
});
```

What I think happens is that when `a` is updated, it calls `mark` recursively down the tree starting at `a` resulting in:
```
a._pending = 1
b._pending = 1
c._pending = 1
d._pending = 2
```
Then, the `sweep` starts at `a`, then goes to `b` where it calls `b`'s `_updater` which recomputes `b` and since its value is the same, it calls unmark on `b`'s subs which are `d`, and in that `unmark` for `d` it decrements `d._pending` bringing it to 1. Then were back in the `sweep` call for `b` and we call `sweep` on all of `b`'s subs, which is `d`. So then in the `sweep` call for `d` the `if (--signal._pending === 0) {` condition decrements `d._pending` to 0 and evaluates to `true`, which causes `d` to recompute unnecessarily.

At least that's what I think is happening from reading the code. I haven't checked it with a debugger.

The issue seems to be related to the fact that both `unmark` and `sweep` can decrement the `_pending` value, and a node that returns the same value can both `unmark` its subs and then `sweep` its subs, which can double decrement their `_pending` value, which I would think is unintended, but I'm not quite sure, I'm still trying to wrap my head around this code.

Is the intended mark and sweep algorithm a standard one that I could read more about somewhere, or is the specific algorithm that is used somewhat unique?

I haven't thought too much about a solution to this issue yet, since I figured those more familiar with the code and the intended algorithm would have a better idea of what to change, but I'd be open to working on a solution if help in that area is wanted.

marvinhagemeister:
This replaces the existing invalidation algorithm with a much simpler one that always recurses upwards from bottom to top. It avoids a bunch of ordering issues that we were running into with the previous algorithm and it's shorter too.

Also kudos to @jviide for sending me various gists that made it click for me!

Fixes #124 .

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -1,6 +1,5 @@
 /** This tracks subscriptions of signals read inside a computed */
 let currentSignal: Signal | undefined;
-let commitError: Error | null = null;
 
 const pending = new Set<Signal>();
 const effects = new Set<Signal>();
@@ -87,14 +86,11 @@ export class Signal<T = any> {
 			// this is the first change, not a computed and we are not
 			// in batch mode:
 			if (isFirst && batchPending === 0) {
-				effects.forEach(signal => activate(signal));
-				pending.forEach(signal => (signal._dirty = false));
-				pending.clear();
-				if (commitError) {
-					const err = commitError;
-					// Clear global error flag for next commit
-					commitError = null;
-					throw err;
+				try {
+					effects.forEach(signal => activate(signal));
+				} finally {
+					pending.forEach(signal => (signal._dirty = false));
+					pending.clear();
 				}
 			}
 		}
@@ -148,25 +144,6 @@ function mark(signal: Signal) {
 	}
 }
 
-function sweep(subs: Set<Signal<any>>) {
-	subs.forEach(signal => {
-		// If a computed errored during sweep, we'll discard that subtree
-		// for this sweep cycle by setting PENDING to 0;
-		if (signal._dirty) {
-			signal._dirty = false;
-
-			// if (signal._isComputing) {
-			// throw Error("Cycle detected");
-			// }
-
-			// signal._isComputing = true;
-			signal._updater();
-			// signal._isComputing = false;
-			sweep(signal._subs);
-		}
-	});
-}
-
 function subscribe(signal: Signal<any>, to: Signal<any>) {
 	signal._active = true;
 	signal._deps.set(to, to._version);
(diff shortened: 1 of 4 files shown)


packages/preact/src/index.ts
```diff
@@ -128,7 +128,7 @@ function Text(this: ComponentType, { data }: { data: Signal }) {
 
 		// Replace this component's vdom updater with a direct text one:
 		currentUpdater!._updater = () => {
-			(this.base as Text).data = s._value;
+			(this.base as Text).data = currentSignal.value.value;
```
developit, line 131: This causes `true` to render as the text "true", instead of using the logic on L137 to make it match VDOM semantics.

marvinhagemeister: Looks like we're missing a test case there

packages/core/src/index.ts
```diff
@@ -103,18 +106,16 @@ export class Signal<T = any> {
...
 		return (shouldUnmark: boolean, shouldCleanup: boolean) => {
-			if (shouldUnmark) this._subs.forEach(unmark);
-
 			// Any leftover dependencies here are not needed anymore
 			if (shouldCleanup) {
 				// Unsubscribe from dependencies that were not accessed:
-				oldDeps.forEach(dep => unsubscribe(this, dep));
+				oldDeps.forEach((_, dep) => unsubscribe(this, dep));
 			} else {
-				// Re-subscribe to dependencies that were not accessed:
```
kidonng, line 116: This seems to accidentally reverting https://github.com/preactjs/signals/pull/129/files#diff-9a4ceebe7c6f86856371906c3f061d3b56b7457022b05179884a113e7ced67e8R116

marvinhagemeister: Good catch!

marvinhagemeister:
Changed the implementation a bit to track dependencies in arrays instead of `Map` because the `Map` version was noticeably slower. With the switch to arrays it's on par with the code in `main` and in my crude tests slightly faster (about ~100ms) when stress testing.

There are some remaining tests in the adapters that are currently failing that I need to address.

marvinhagemeister:
Ok, doing some more checks the code in this PR is actually quite a bit faster than in `main`. In raw stress testing creation of computed signals and triggering one update it's about 1.4x faster

marvinhagemeister:
Closing as this PR will be replaced by #136 .

PR #133, opened 2022-09-13
fix: Work-around for relative README links

rschristian:
## Issue

Relative links to the repo root work great for our subpackages while browsing on GitHub, but this causes issues when published to NPM.

NPM's combination of the link with the repo URL unfortunately allows users to escape platform-specific path parameters. See the following:

```
Source Link: `../../README.md`

Correct Link:  `<repo>/tree/HEAD/README.md`
NPM's Link:    `<repo>/README.md`
```

~~Might be a bit too hopeful on my part for how intelligently NPM would handle this, but a disappointment nonetheless.~~

## Summary

The fix here (to which all credit goes to @kidonng) is just to specify the directory of the package in our `package.json` for each subproject.

NPM can figure out the links from there.


packages/preact/package.json
```diff
@@ -9,7 +9,8 @@
 	],
 	"repository": {
 		"type": "git",
-		"url": "https://github.com/preactjs/signals"
+		"//": "This is a poor workaround for github.com/npm/feedback/discussions/768. Only way to get working relative links on NPM & GitHub",
+		"url": "https://github.com/preactjs/signals/blob/HEAD"
```
kidonng, line 13: You could use `/tree/` instead of `/blob`, so it works both for the homepage and files (GitHub will redirect `/tree/` to `/blob/` when necessary).

rschristian: Interesting, indeed it does!

rschristian: Oh wait, nope. NPM will not handle that correctly.

rschristian: https://www.npmjs.com/package/@rschristian/signals

https://www.runpkg.com/?@rschristian/signals@0.0.12/package.json

Repository URL is set to `https://github.com/rschristian/signals/tree/HEAD` yet the relative link on NPM is to `https://github.com/rschristian/signals/README.md#guide--api`

kidonng: I see, npmjs.com probably does some normalization and strips `/tree/<whatever>/`.

rschristian: Seemingly yes. Keeps `/blob/<whatever>` though, which is a bit odd.

kidonng: What if you use the shorthand like `"repository": "preactjs/signals"`? I looked at some other packages like [xo](https://www.npmjs.com/package/xo) and it does not seem to have this issue.

rschristian: `xo` doesn't look to contain any relative links.

That format, at least in the past, wouldn't work. Likely still won't.

kidonng:
@rschristian I just realized this is for sub directories, while I was looking at the root README all the time 😅

This should do the trick:

```
"repository": {
   "type": "git",
   "url": "https://github.com/preactjs/signals",
   "directory": "packages/preact"
}
```

I published a demo package to [npm](https://www.npmjs.com/package/@kidonng/test-package), you can [see the contents here](https://unpkg.com/browse/@kidonng/test-package@0.0.1/).

rschristian:
Ah shoot, yep that seems to work.

Someone from NPM once told me that was just metadata and not used for anything on their end. Didn't even think to check that.

marvinhagemeister (approved):
Nice 🎉

PR #136, opened 2022-09-13
Implement core without Sets, Maps, Arrays etc.

jviide:
This pull request rewrites the core to avoid allocating Sets, Maps and Arrays on the fly. The implementation uses linked lists of nodes that contain both the source (signal) and target (computed/effect) of dependency-subscription pairs. The aim was to facilitate constant-time subscribe/unsubscribe operations without sets.

Currently this breaks the Preact/React integrations, as the core internals have changed quite a bit.

## Very lazily computed signals

Computed signals evaluate their values lazily whenever their `.value` is accessed or `.peek()` is called.

Computed signals also lazily subscribe to be notified when their dependencies change _only_ when the computed itself already has subscribers. This is done so that computed signals that aren't currently (direct or indirect) dependencies of some effects or other computed signals can be garbage collected when they go out of scope.

When value of a subscribeless computed signal is accessed the computed signal always checks whether its dependencies have changed. To speed this process up, signals also keep track of their own "version numbers" that change whenever their value changes.  Computed signals then store (in the linked list nodes) last version numbers they've seen from their dependencies, and use that info to estimate whether their computation function has to be re-evaluated.

> Note that the computed signals _could_ store the actual dependency values instead of version numbers. However I went with the version numbers, because a single dependency value can take an arbitrary amount of memory. As computed signals are lazily evaluated they could theoretically hang on to those already out-of-date dependency values indefinitely before the computed is evaluated again. 

When a computed has subscribers it gets notifications about changed dependencies as per usual, and the recomputation process can be short-circuited when the computed signal's `._valid` flag is `true`.

Every signal value change increments a global version number. Computed signals remember what the global version number was when their computed value was last validated. This info is then used to fast-track `.value`/`.peek()` calls when nothing has changed globally.

## Error handling

If a computed signal value function throws, the error is stored and subsequent `.value`/`.peek()` calls throw the same error without re-evaluating the compute function until one of the computed signal's dependency values changes.

If a computed value depends on itself (directly or indirectly) a "cycle detected" error is raised inside the compute function. To prevent runaway recursive effects, the top level batch handler tracks how many operation batches it has had to process consecutively. After 100 iterations all 'signal.value = ...` calls throw a "cycle detected" error.

When an effect in a batch handler throws an error it's percolated to the calling context that started the topmost batch handler. If multiple effects fail inside one batch handling loop then only the first error is percolated to the top.

## Notes about avoiding sets

The main use case for using Sets instead the original implementation was to avoid subscribing to the same dependency twice inside an effect or a computed signal's value function. Arrays can be used instead of Sets, but for then avoiding duplicate subscriptions requires checking the array for duplicates, which is an O(n) operation.

This PR tried to solve this with the linked lists, by making signals keep track of the last evaluation context in the stack that has accessed their value. Signals can then can quickly check whether they've already been inserted to the evaluation context's dependencies. This way the check is constant-time.

Signal's `._node` property keeps track of the last linked list node in the evaluation stack where the signal is the dependency. This way `_node.target` value can be used for the evaluation context check described above, but also for recycling nodes.

Those same linked list nodes are also used to remember `node.signal`'s `._node` values after entering a new evaluation context, and then for restoring the `._node` values when exiting from the evaluation context.

Batching is done with a linked list of effects that should be run in the next batch iteration. To avoid memory churn Effect instances themselves act as nodes in that linked list.

## Benchmark

This (extremely simplistic, should be run with d8 etc.) benchmark:

```ts
import * as core from "@preact/signals";
import * as setless from "./the-new-core";

{
  const count = core.signal(0);

  console.time("core");

    count.value++;
    double.value;
  }

  console.timeEnd("core");
}

{
  const count = setless.signal(0);
  const double = setless.computed(() => count.value * 2);

  console.time("setless");

    count.value++;
    double.value;
  }

  console.timeEnd("setless");
}
```

I get on my machine:

```sh
$ npx ts-node benchmark.ts
core: 6.385s
setless: 458.465ms
```

### DIFF
--- a/packages/core/test/signal.test.tsx
+++ b/packages/core/test/signal.test.tsx
@@ -215,12 +215,40 @@ describe("computed()", () => {
 		expect(() => a.value).to.throw(/Cycle detected/);
 	});
 
+	it("should not allow a computed signal to become a direct dependency of itself", () => {
+		const spy = sinon.spy(() => {
+			try {
+				a.value
+			} catch {
+				// pass
+			}
+		})
+		const a = computed(spy);
+		a.value;
+		expect(() => effect(() => a.value)).to.not.throw();
+	});
+
 	it("should detect deep dependency cycles", () => {
 		const a: Signal = computed(() => b.value);
 		const b: Signal = computed(() => c.value);
 		const c: Signal = computed(() => d.value);
 		const d: Signal = computed(() => a.value);
 		expect(() => a.value).to.throw(/Cycle detected/);
+	});
+
+	it("should store failures and recompute after a dependency changes", () => {
+		const a = signal(0);
+		const spy = sinon.spy(() => {
+			a.value;
+			throw new Error();
+		});
+		const c = computed(spy);
+		expect(() => c.value).to.throw();
+		expect(() => c.value).to.throw();
+		expect(spy).to.be.calledOnce;
+		a.value = 1;
+		expect(() => c.value).to.throw();
+		expect(spy).to.be.calledTwice;
 	});
 
 	it("should conditionally unsubscribe from signals", () => {
--- a/packages/core/test/signal.test.tsx
+++ b/packages/core/test/signal.test.tsx
@@ -161,6 +161,24 @@ describe("effect()", () => {
 		expect(fn).to.throw(/Cycle detected/);
 	});
 
+	it("should allow disposing the effect multiple times", () => {
+		const dispose = effect(() => undefined);
+		dispose();
+		expect(() => dispose()).to.throw;
+	});
+
+	it("should throw when disposing a running effect", () => {
+		const a = signal(0);
+		const dispose = effect(() => {
+			if (a.value === 1) {
+				dispose();
+			}
+		});
+		expect(() => {
+			a.value = 1;
+		}).to.throw("Effect still running");
+	});
+
 	it("should not run if it's first been triggered and then disposed in a batch", () => {
 		const a = signal(0);
 		const spy = sinon.spy(() => a.value);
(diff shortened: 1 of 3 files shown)


marvinhagemeister:
Really like where this is going 🎉

marvinhagemeister:
I like even more where this is going 🚀

packages/core/test/signal.test.tsx
```diff
@@ -251,7 +404,7 @@ describe("computed()", () => {
 			compute.resetHistory();
 
 			a.value = 4;
-
+			e.value;
```
developit, line 407:
```suggestion
			d.value;
```

marvinhagemeister (approved):
This is a fantastic piece of engineering!! Really impressed with the results and the speed improvements 👍

packages/core/test/signal.test.tsx
```diff
@@ -160,6 +160,213 @@ describe("effect()", () => {
 
 		expect(fn).to.throw(/Cycle detected/);
 	});
+
+	it("should allow disposing the effect multiple times", () => {
```
cevr, line 164: sorry for a random comment from a random person, was just reading through this interesting code

but this test appears to be showing that disposing the effect multiple times is not allowed

jviide: And thanks a bunch for looking through the code! The [current main branch version](https://github.com/preactjs/signals/blob/main/packages/core/src/index.ts) has been iterated on further quite a bit, so I recommend checking it out. In both cases apologies for the code's readability, there are some comments sprinkled there but they probably convey all the tricks there 😅

jviide: Argh, update: I meant to say "they probaby **don't** convey all the tricks there".

PR #139, opened 2022-09-13
Add release script

marvinhagemeister:
This PR adds a release script to `package.json`. Reason being that with `pnpm` we cannot do just `pnpm changeset publish` because it stores the versions in the lockfile. That lockfile needs to be updated with every release and has led to a few annoyances in the past.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,9 @@
 		"docs:build": "cd docs && pnpm build",
 		"docs:preview": "cd docs && pnpm preview",
 		"ci:build": "pnpm build && pnpm docs:build",
-		"ci:test": "pnpm lint && pnpm test"
+		"ci:test": "pnpm lint && pnpm test",
+		"release": "pnpm changeset version && pnpm install && git add -A && git commit -m 'Version Packages' && changeset tag && pnpm publish -r",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -21,7 +21,8 @@
 		"docs:build": "cd docs && pnpm build",
 		"docs:preview": "cd docs && pnpm preview",
 		"ci:build": "pnpm build && pnpm docs:build",
-		"ci:test": "pnpm lint && pnpm test"
+		"ci:test": "pnpm lint && pnpm test",
+		"release": "pnpm changeset version && pnpm install && git add -A && git commit -m 'Version Packages' && changeset tag && pnpm publish -r"
```
rschristian, line 25:
```suggestion
		"release": "pnpm install && pnpm changeset version && git add -A && git commit -m 'Version Packages' && pnpm changeset tag && pnpm publish -r"
```

JoviDeCroock (approved):

package.json
marvinhagemeister: The install command has to happen after `changeset version` which is the root cause of the problems we are having. `pnpm` mirrors the package versions in all`package.json` files in their lockfile.

rschristian: Ohh, gotcha.

Never used pnpm, but that looked odd to me.

PR #143, opened 2022-09-14
Changing batch to use a single flag instead of a counter

Pauan:
As discussed in #104, I'm going to try and split the improvements into multiple smaller PRs.

All of the tests pass, but this does slightly change the behavior, so I hope it doesn't break anybody's code.


marvinhagemeister (approved):
Awesome, this is great! Thank you for the PR 👍

PR #144, opened 2022-09-14
Add a basic devcontainer

jviide:
This pull request adds a basic devcontainer setup:
 * pnpm installed
 * Chromium installed for running tests
 * ESLint and Prettier extensions for VSCode

I went with Alpine, as I have the most experience with it and it had Chromium etc. nicely packaged for both x86_64 and aarch64. Tested with a M1 based Mac.


marvinhagemeister (approved):
Awesome, thanks a bunch for the PR 💯

PR #151, opened 2022-09-14
Use GitHub aware formatter for changesets

marvinhagemeister:
Currently our changeset action generates release notes in the format of

```txt
<hash> - <msg>
```

That works but it would be much nicer if it could also link the PRs and tag the contributor automatically.

This PR switches over to the same formatter that we used with WMR.


.changeset/config.json
```diff
@@ -1,6 +1,6 @@
 {
 	"$schema": "https://unpkg.com/@changesets/config@2.1.1/schema.json",
-	"changelog": "@changesets/cli/changelog",
+	"changelog": "@changesets/changelog-github",
```
rschristian, line 3: Do we also need to specify a repo? 

https://github.com/preactjs/wmr/blob/6a9d7768baee41cf90ef30f2c3cfb5c1a207c78e/.changeset/config.json#L3-L6

https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#changelog-false-or-a-path

marvinhagemeister: You're right on. That needs to be added.

JoviDeCroock (approved):

PR #158, opened 2022-09-16
Support auto-disposing nested effects

jviide:
This pull request adds support for auto-disposing nested effects. With the following setup:

```ts
const a = signal("a");
const b = signal("x");
const e = effect(() => {
  console.log("outer effect", a.value);
  effect(() => {
    console.log("inner effect", a.value + b.value);
  });
});

// Console: outer effect a
// Console: inner effect ax
```

Now if we run the outer effect again the previous inner effect is auto-disposed:

```ts
a.value = "b";
// Console: outer effect b
// Console: inner effect bx
```

If we only update the `b` signal only the latest version of the inner effect runs:

```ts
b.value = "y"
// Console: inner effect by
```

When we dispose the outer effect then the inner effect is also disposed:

```ts
e();
a.value = "c";
// Nothing in the console
```

Note that only immediate containing effects are tracked: For example creating an effect inside an computed inside an effect doesn't dispose the innermost effect when the outermost effect is disposed.

### DIFF
--- a/.changeset/heavy-turkeys-relate.md
+++ b/.changeset/heavy-turkeys-relate.md
@@ -1,0 +1,5 @@
+---
+"@preact/signals-core": minor
+---
+
+Add support for auto-disposing nested effects
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -58,8 +58,8 @@ function endBatch() {
 		batchIteration++;
 
 		while (effect !== undefined) {
-			const next: Effect | undefined = effect._nextEffect;
-			effect._nextEffect = undefined;
+			const next: Effect | undefined = effect._nextBatchedEffect;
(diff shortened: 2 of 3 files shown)


marvinhagemeister (approved):
This is great!!

WebReflection:
> creating an effect inside an computed inside an effect doesn't dispose the innermost effect when the outermost effect is disposed.

usignal used to have WeakRef to deal with computed but it recently got moved for various reasons ... this scenario though seems like OK as once the inner computed gets collected its inner most effect should get collected too.

it's an interesting case though, 'cause the inner effect if run *when* the outer runs we have a reference, but if the computed is accessed conditionally we might have issues, but as long as the effect internals don't leak on purpose outside the effect I believe everything should be covered 👍

PR #160, opened 2022-09-16
Implement Effect, Computed and Signal classes as ES5 classes

jviide:
This pull request implements the Effect, Computed and Signal classes as ES5 classes.

Added a workaround for a weird performance cliff that I encountered with Node.js 18.9.0. The workaround itself is in commit 64d0ab0844a4d50ee09d564b8748127e8a957049. For some reason the TypeScript compiled version was faster than the hand-rolled ES5 classes when running this benchmark:

```ts
import * as core from "@/preact/signals-core";

{
  const count = core.signal(0);

  console.time("core");

    count.value++;
    double.value;
  }

  console.timeEnd("core");
}
```

In the end it turned out that TypeScript wrapped the Computed prototype definitions into an IIFE. Adding that same IIFE to the hand-rolled ES5 version boosted the performance back to TypeScript-compiled levels 🫠

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,10 +29,11 @@
 	"author": "",
 	"license": "MIT",
 	"devDependencies": {
-		"@babel/core": "^7.18.10",
-		"@babel/preset-env": "^7.18.10",
+		"@babel/core": "^7.19.1",
+		"@babel/preset-env": "^7.19.1",
 		"@babel/preset-react": "^7.18.6",
 		"@babel/preset-typescript": "^7.18.6",
+		"@babel/plugin-transform-typescript": "^7.19.1",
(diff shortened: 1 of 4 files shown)


marvinhagemeister:
The test failure is caused by a bug in babel's typescript plugin. Filed an issue there: https://github.com/babel/babel/issues/14945

marvinhagemeister:
@jviide This should be ready to be merged once rebased. CI passes locally 🎉

gamedevsam:
Doesn't this reveal a problem with Node v18 though? Seems very strange and not at all what I'd expect. @mcollina this make sense at all to you?

PR #162, opened 2022-09-16
Add Signal.prototype.valueOf

developit:


packages/core/src/index.ts
```diff
@@ -224,6 +224,10 @@ export class Signal<T = any> {
 		return effect(() => fn(this.value));
 	}
 
+	valueOf(): T {
+		return this.value;
+	}
```
marvinhagemeister, line 229: Let's add some tests for this 👍

marvinhagemeister (approved):

PR #165, opened 2022-09-16
Dispose computed signal when enclosing scope gets disposed

jviide:
This pull request modifies the autodisposal logic by also disposing computed signals when the enclosing scope (effect / computed signal) gets disposed.

After a computed signal has been disposed trying to access its value always throws an error ("Computed disposed").

If the disposal happens while the computation function is being evaluated then the evaluation runs to the end like normal, the computation gets disposed properly and it throws the same error as above.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -317,22 +317,38 @@ function cleanupSources(target: Computed | Effect) {
 	target._sources = sources;
 }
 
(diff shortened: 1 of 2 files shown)


jviide:
On further thought how we handle computed signals is probably fine as it is. As it is you have to a bit out of your way to create a resource leak, so I'm closing this PR 🙂

PR #170, opened 2022-09-17
Allow disposing currently running effects

jviide:
This pull request adds the support for disposing currently running effects. The reasoning for this is that it's convenient, and ends up saving a couple of bytes in some bundle versions.

Disposing the effect while running just delays the full disposal until the run has finished.


marvinhagemeister (approved):
Niice

marvinhagemeister:
@jviide feel free to add a changeset for this. I think it's worthy enough to be part of the release notes 💯

PR #171, opened 2022-09-18
Refactor Preact integration

jviide:
This pull request refactors the Preact integration to modestly reduce the size:
 * Modify property change listeners to use the signals' own change tracking to track whether the signal that the property should listen has changed.
 * Keep the state in previously kept in WeakSets and a WeakMaps Component's own props. The updater is stored in `_updater` and `hasHookState`, `hasPendingUpdate` and `hasComputeds` are tracked in the bitset `_updateFlags`. The names get mangled to `_$u` and `_$f`, respectively.

### DIFF
--- a/mangle.json
+++ b/mangle.json
@@ -56,6 +56,9 @@
       "$_subscribe": "S",
       "$_unsubscribe": "U",
       "preact": "",
+      "$_signal": "s",
+      "$_updater": "_$u",
+      "$_updateFlags": "_$f",
       "$_updaters": "U"
     }
   }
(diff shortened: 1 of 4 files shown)


marvinhagemeister (approved):
This is really nice!

mangle.json
```diff
@@ -56,6 +56,9 @@
       "$_subscribe": "S",
       "$_unsubscribe": "U",
       "preact": "",
+      "$_signal": "s",
+      "$_updater": "_$u",
```
developit, line 60: I think these need to be double underscores so they don't get reminified?

jviide: Oh, right, sorry about that 🙂 Opened #176 to fix it.

PR #173, opened 2022-09-18
Add batch tests

jviide:
This pull request adds some tests for `batch()`, focusing on how `batch()` handles different return values and thrown errors from batch callbacks and effects.


marvinhagemeister (approved):
Nice, always love seeing more tests 👍

PR #175, opened 2022-09-18
Add favicon to demo html

marvinhagemeister:
It's the same favicon from preactjs.com . Will get rid of the console error browsers print.


JoviDeCroock (approved):

developit:
@marvinhagemeister fun hack:

```
<link rel=icon href="data:,">
```

marvinhagemeister:
@developit That's clever! I was googling for "empty favicon" but all I got was a bunch of favicon generators.

PR #176, opened 2022-09-18
Don't mangle new Component helper props

jviide:
Fix mangling of new Componen properties introduced in #171.

### DIFF
--- a/mangle.json
+++ b/mangle.json
@@ -57,8 +57,8 @@
       "$_unsubscribe": "U",
       "preact": "",
       "$_signal": "s",
-      "$_updater": "_$u",
-      "$_updateFlags": "_$f",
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
I'm curious: Did this trigger a bug that we didn't catch with our test suite?

jviide:
I don't think so. Currently we don't have a test for this.

The reasoning for the naming is that I didn't want the mangling to accidentally rename `_updater` and `_updateFlags` to names that are already being used by Preact itself. Is this even something that I should worry about? 😄

marvinhagemeister:
Fair point! Haven't thought about potential name clashes with Preact itself. That makes sense.

PR #177, opened 2022-09-19
Fix current updater

developit:


marvinhagemeister (approved):

marvinhagemeister:
@developit The alternative solution in #179 looks a lot cleaner to me. Which one do you prefer?

marvinhagemeister:
Closing, because of the alternative approach in #179 being merged.

PR #178, opened 2022-09-19
Hide Signal and Computed functions from output .d.ts files

jviide:
This pull request adds `/** @internal */` annotations above the Signal and Computed functions to hide them from the build .d.ts files.


marvinhagemeister (approved):
Good catch!

PR #183, opened 2022-09-19
Implement effect cleanups

jviide:
This pull request implements effect cleanup functions. An effect callback can return a function that's then run the next time the effect either gets rerun or disposed. This pull request also converts the `useSignalEffect` hook for the Preact and React integrations to use effect cleanups.

The `effect(compute: () => void)` signature is modified to `effect(compute: () => unknown)`:

```ts
const s = signal(0);

const dispose = effect(() => {
  if (s.value === 0) {
    effect(() => () => console.log("hello"));
  } else {
    effect(() => () => console.log("what a"));
  }

  return () => console.log("world");
});

s.value++; // Prints "hello" & "world"
dispose(); // Prints "what a" & "world"
```

Apart from testing basic functionality, the tests try to cover several corner cases and specific behaviors:
 * The cleanup functions of nested effects are run before the parent effect's cleanup.
 * When an effect or computed signal is running its nested effect cleanups, and one of the cleanups throws, the rest of the cleanups are still completed and the error is only then thrown upwards.
 * If one of effect's nested cleanups throws the effect's own cleanup function still gets called.
 * If an effect's own cleanup or one of its nested cleanups throw, the effect becomes essentially disposed. The error is thrown up the stack.
 * If a computed signal's nested effects fail when the computed value is being revalidated, all of the computed signal's dependency links are severed. Accessing the computed's value from thereon throws the error.
 * If an effect throws an error then its cleanup function is assumed to be undefined.
 * Effects can be disposed and retriggered during their own cleanups or nested cleanups.

### DIFF
--- a/packages/react/src/index.ts
+++ b/packages/react/src/index.ts
@@ -169,16 +169,8 @@ export function useSignalEffect(cb: () => void | (() => void)) {
 	callback.current = cb;
 
 	useEffect(() => {
-		let cleanup: (() => void) | undefined;
 		return effect(() => {
-			if (cleanup) {
-				cleanup();
-				cleanup = undefined
-			}
-			const result = callback.current();
-			if (typeof result == 'function') {
-				cleanup = result
-			}
-		})
+			return callback.current();
+		});
 	}, []);
 }
--- a/.changeset/tidy-buses-do.md
+++ b/.changeset/tidy-buses-do.md
@@ -1,0 +1,16 @@
+---
+"@preact/signals-core": minor
+"@preact/signals": minor
+"@preact/signals-react": minor
+---
+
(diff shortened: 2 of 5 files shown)


packages/react/src/index.ts
```diff
@@ -169,16 +169,8 @@ export function useSignalEffect(cb: () => void | (() => void)) {
...
-			if (cleanup) {
-				cleanup();
-				cleanup = undefined
-			}
-			const result = callback.current();
-			if (typeof result == 'function') {
-				cleanup = result
-			}
-		})
+			return callback.current();
```
marvinhagemeister, line 173: I like that `useSignalEffect` got a lot simpler!

marvinhagemeister (approved):
This is awesome!

PR #190, opened 2022-09-19
Proposal: Implement Vue-style effect cleanups

jviide:
This pull request suggest the following middle point between having and not havig autodisposed effects. It also modifies effect cleanups introduced in #183, allowing scheduling multiple effect cleanups during the same effect run:

```ts
import { signal, effect } from "@preact/signals-core";

effect((cleanup) => {
  cleanup(() => {
    console.log("this will run the next time the effect rerun or disposed");
  });
  cleanup(() => {
    console.log("ditto");
  });
});
```

Effects aren't parented to their contexts by default anymore, and there is no autodisposal. It's pretty easy to achieve a similar effect:

```ts
import { signal, effect } from "@preact/signals-core";

effect((cleanup) => {
  cleanup(effect(() => {
    console.log("this effect will be autodisposed, sort of");
  }));
});
```

This version avoids autodisposal's "spooky action by distance" (accidentally binding an effect's lifetime to another one) by making the disposals explicit, while getting most of autodisposal's benefits. It also requires a bit less code 🙂

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -6,11 +6,9 @@ function cycleDetected(): never {
 const RUNNING = 1 << 0;
 const NOTIFIED = 1 << 1;
 const OUTDATED = 1 << 2;
-const DISPOSED = 1 << 3;
-const HAS_ERROR = 1 << 4;
-const IS_EFFECT = 1 << 5;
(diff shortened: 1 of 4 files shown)


jviide:
Based on Slack discussion we won't be going with this approach for now. Superceded by #191.

PR #209, opened 2022-09-24
Optimize dependency value change checks by allowing earlier exists from the loop

jviide:
This pull request modifies the dependency version change check used by computed & effect. Before the change each dependency's `._refresh()` was called before checking the dependency's version number. After this change the version number is checked first, then `._refresh()` is called, and then the version number is checked again. In practice this allows the dependency checks break the loop earlier, avoiding calls to `_refresh()` for a) plain signals that have changed and b) computed values that get repeated reads.

The following benchmark demonstrates the impact in a scenario that contains both a) and b):

```ts
import * as core from "@preact/signals-core";

const count = core.signal(0);

console.time("core");

count.value++;
double.value;
}

console.timeEnd("core");
```

The results before the change:
```
core: 2.616s
```

And after the change:
```
core: 2.196s
```

### DIFF
--- a/.changeset/tall-onions-deny.md
+++ b/.changeset/tall-onions-deny.md
@@ -1,0 +1,5 @@
+---
+"@preact/signals-core": patch
+---
+
+Optimize dependency value change checks by allowing earlier exists from the loop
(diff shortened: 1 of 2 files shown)


marvinhagemeister (approved):
🎉

JoviDeCroock (approved):

Issue #155, opened 2022-09-15
signals-react doesn't work in react production build


dackom:
The basic example is not working on production build in react:

```
const count = useSignal(1);
return (
<div>
   {count}
   <button onClick={() => count.value++}>INC</button>
</div>
);
```

Versions:
```
    "@preact/signals-react": "^1.0.2",
    "react": "^18.2.0",
    "react-dom": "^18.2.0",
```

Building with vite 3.1.0

developit:
@marvinhagemeister I wonder if we missed the publish for the react integration too, like in #140?

marvinhagemeister:
@developit Can confirm that it's not a bad publish, but a bug.

An-Huynh-Certivity:
I am having this exact same issue.

```
import { effect, signal } from '@preact/signals-react'

function Test() {
  const foo = signal<boolean>(false)
  effect(() => {
    console.log(foo.value)
  })

  const counter = signal(0)
  effect(() => {
    console.log(counter.value)
  })

  return (
    <>
      <div>
        <div>{foo.value ? 'true' : 'false'}</div>
        <button
          className="bg-gray-400"
          onClick={() => {
            foo.value = !foo.value
          }}
        >
          invert
        </button>
      </div>
      <div>
        <div>{counter.value}</div>
        <button
          className="bg-gray-400"
          onClick={() => {
            counter.value += 1
          }}
        >
          increment
        </button>
      </div>
    </>
  )
}

export default Test
```

with the same package.json as the original post.

On invert, the console logs:
```
true
false
0
```

On increment, the console logs:
```
1
false
0
```

mohammadraufzahed:
any progress?

jmeistrich:
For what it's worth I haven't found `ReactCurrentOwner` to be dependable. In my testing in production it seems to always be set to the root node, and in React Native it's even less dependable. So for Legend-State's implementation I used only `ReactCurrentDispatcher` to track lifecycle, which does work in production and React Native.

I'm interested to see if you're able to find a workable solution with `ReactCurrentOwner` (or something else) but it may not be possible.

developit:
@jmeistrich are you ending up with two copies of React?

jmeistrich:
@developit No, I don't think so. I've been testing in three different production apps (React 17, 18, React Native) and I don't see consistent `ReactCurrentOwner` behavior. In production sometimes it's set to the root node, sometimes it's always null (I don't remember which was which).

[...]

PR #215, opened 2022-09-26
Change mapping updaters to owners to map to the reducer, fixing not working in production

jmeistrich:
As I mentioned in https://github.com/preactjs/signals/issues/155, tracking the current owner is unstable and doesn't work in production. I think the purpose of that was to keep a unique updater for each component. I found that can also be done by just using the rerender callback as the key for the map since it is consistent and unique to each component's render.

This change also changes the invalid hook check to just "Invalid" which works in React Native too (its hook warnings are slightly different).

The tests still pass, and this method is working well in Legend-State (https://github.com/LegendApp/legend-state/blob/main/src/react/enableLegendStateReact.ts#L70) in production and React Native.

### DIFF
--- a/packages/react/src/index.ts
+++ b/packages/react/src/index.ts
@@ -15,7 +15,7 @@ import {
 	Signal,
 	type ReadonlySignal,
 } from "@preact/signals-core";
-import { Effect, ReactOwner, ReactDispatcher } from "./internal";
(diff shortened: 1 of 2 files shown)


marvinhagemeister (approved):
This is awesome! Thank you so much for the PR 🙌

Issue #213, opened 2022-09-25
Allow a Signal to track multiple values using a bitmap (32bit integer)


eddyw:
Hey 👋 

While my git command is unusable thanks to Xcode update, I thought to open a discussion about something I'm trying to hack together in my local signals clone repo.

From my understanding, a `Node` is sort of like a subscription, right? so it uses `version` instead of the actual `value` of the signal instance to know when a signal's value has changed. 

Based on this idea, we could use a bitmap (a 32bit value) to track changes in 32 values. Each bitfield represents the position of the property, call it `value0, value1, ..., value31` -> `0b1...111`.

The idea is more or less this:
```ts
computed(() => {
  signal.value0; // Pos: 1 << 0
  signal.value1; // Pos: 1 << 1
  signal.value5; // Pos: 1 << 5
});

// (Pseudo) What roughly happens:

let fields = 0;  // Observed/watched fields
let version = 1; // Current version

fields |= 1 << 0; // <- get value0
fields |= 1 << 1; // <- get value1
fields |= 1 << 5; // <- get value5

computedNode.version = version;
computedNode.fields = fields; // 0b100011 Observed/watched fields by this `computed`
fields = 0;
```
This means, the current version is `1` and the signal values accessed are `value0`, `value1`, and `value5` inside the computed. In other words, 1 single node is tracking changes in these 3 signal accessed fields using a single 32bit integer.

Now, when a value change occurs:
```ts
signal.value6 = "foo";

// (Pseudo) What roughly happens:

version += 1;
fields |= 1 << 6; // <- guess within batch could set multiple fields

if (
  computedNode.version !== version &&
  computedNode.fields & fields // it's '&' bitwise op, not '&&'
) computedNode.recompute()
```

In this particular case, `value6` wouldn't cause the computed node to re-compute because - while the signal's version has changed - the observed values (0, 1, and 5) have not changed. 

It's also fairly cheap to check which fields changed using a bitwise operator and v8 (at least) does a good job when you work only with 32bit integers (as long as the value is never set to float by accident).

For a bit of context, I have a factory that creates constructors, it adds getters/setters to the prototype which are wrapped in a signal. While signals have a relatively small memory footprint, I think this could be improved a lot. I use a similar approach to create 1 signal-like for every 32 properties. At best a single instance of a class has less than 32 properties so it can be used to observe all of them instead of creating multiple subscriptions for every property that's accessed in a computed for instance.

[...]

PR #217, opened 2022-09-26
Proposal: Multi-value Signal using bitmaps (bit arrays)

eddyw:
Hi 👋, this is a PoC of #213 😄 (all tests still passing 😅)

# Description

The basic idea of this PR is demonstrate how Signals could be improved to keep track of multiple values (up to 32) using 1 single node to reduce memory footprint and possibly improve performance when multiple (related) values are accessed within a `computed` or `effect`.

In this PoC, assume a Signal can have - instead of value - properties `signal[0] (alias for signal.value)` to `signal[31]`. Just because numeric properties seem to be accessed faster than string properties (it could have been `signal.value0 .. signal.value31` but it's not relevant).

For every `get`, we keep track of the index - in a bit array - of the accessed property:
```ts
get(this: Signal) {
  // ...
  node._version = this._version;
  node._fields |= 1 << i; // i -> 0..31 -> signal[0]..signal[31]
  // ...
  return this._value[i];
}
```

And every `set` operation checks whether the node is watching the property whose value is changed:
```ts
set(this: Signal, value) {
  // ...
  this._value[i] = value; // i -> 0..31
  // ...
  if (node._fields & (1 << i)) node._target._notify(); // < Notify ONLY if property is watched
}
```

With this as base, we can create signals that have multiple values (up to 32) and create a lot less nodes if all the accessed properties within a `computed` belong to the same Signal, for example:

```ts
computed(() => signal[0] + signal[1] + signal[2]);
computed(() => signal[1] + signal[5] + signal[31]);
```

In this particular example, all properties in both `computed` belong to the same signal, so a single node should be created with bitmap set to `fields = (1 << 0) | (1 << 1) | (1 << 2)` (for the first). This means that, for example, if `value[0]` changes, then the second `computed` won't run even though both use the same signal.

Now, the DX is awkward so this PoC doesn't attempt to be a final version but rather to suggest creating an even lower primitive than Signal. A Signal could be an abstraction for this new primitive where it only uses one field `value[0]`.

An example for such abstraction is the `reader` function which I included in the code. Which allows creating signals in this manner:
```ts
const group = reader({
  foo: "",
  bar: "",
});
// group.foo;
// group.bar;
// group._signals = [Signal]
// group._size = 2;
```
( Please check the test cases for `reader` 😄 )

A Signal could look like:
```ts
function signal(value) {
  const s = reader({ value });
  s.peek = function() {
    return s._signals[0][0];
  }
}
```
Of course, this example is just an illustration since better optimizations can be made but I think it conveys the point 😅 

## But Why?

I want to create a factory which creates classes, sort of like this:
```ts
class User extends Factory({
  username: t.String(),
  // ... many other fields
}) {}
```
Where `t.String` is a config object which is converted to a property descriptor with getter/setter which _would_ use a signal. However, this consumes a lot of memory and it's (relatively) slow if multiple related properties are accessed within a `computed`.

I think this could be improved by using a single Signal to watch multiple related properties (well, up to 32). As in the `reader` PoC example, in one of the tests, you can see that 93 properties can be used with just 3 signals instead of defining 93 signals.

My implementation is just a PoC, at its current state I just hope to spawn some discussion around this if it's interesting enough to move forward within preact/signals (Please check the new test cases).

I think this could also be interesting for #4 . I don't think all this logic should belong to `Signal` 😅 to be honest, it was just easier to implement it there for a PoC PR.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -26,6 +26,7 @@ type Node = {
 	_source: Signal;
 	_prevSource?: Node;
 	_nextSource?: Node;
+	_fields: number;
 
 	// A target that depends on the source and should be notified when the source changes.
 	_target: Computed | Effect;
@@ -134,6 +135,7 @@ function addDependency(signal: Signal): Node | undefined {
 			_prevTarget: undefined,
 			_nextTarget: undefined,
 			_rollbackNode: node,
+			_fields: 0,
 		};
 		evalContext._sources = node;
 		signal._node = node;
@@ -176,7 +178,21 @@ function addDependency(signal: Signal): Node | undefined {
 
 declare class Signal<T = any> {
 	/** @internal */
-	_value: unknown;
+	_value: unknown[];
+	0: unknown;
+	1: unknown;
+	2: unknown;
+	3: unknown;
+	4: unknown;
+	5: unknown;
+	6: unknown;
+	7: unknown;
+	8: unknown;
+	9: unknown;
+	10: unknown;
+	11: unknown;
+	// ... up to 👇
+	31: unknown;
 
 	/** @internal */
 	_version: number;
@@ -187,7 +203,7 @@ declare class Signal<T = any> {
 	/** @internal */
 	_targets?: Node;
 
-	constructor(value?: T);
+	constructor(value?: T, ...other: any[]);
 
 	/** @internal */
 	_refresh(): boolean;
@@ -211,8 +227,8 @@ declare class Signal<T = any> {
 }
 
 /** @internal */
-function Signal(this: Signal, value?: unknown) {
-	this._value = value;
+function Signal(this: Signal) {
+	this._value = Array.from(arguments);
 	this._version = 0;
 	this._node = undefined;
 	this._targets = undefined;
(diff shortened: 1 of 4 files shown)


developit:
Hi Eddy! I have a hunch that this may be a bit out-of-scope for signals, since we're focused on providing a single "box" primitive that folks can build interesting things like this on top of. Perhaps it would make sense to explore this in one of the object/wrapper libraries folks have been working on that use Signals under the hood?

marvinhagemeister:
I think this is a fantastic prototype and well worth exploring further! Agree with @developit that it's currently out of scope for signals as we're trying to keep the API as minimal as possible. The PR here would be perfect as a separate package that users can install and try out 👍

jviide:
In the current tracking scheme (based on versions numbers) each of the 32 different values probably need their own version numbers. That's because a signal can have multiple listeners (effects, computeds) that run their computations in different times. For example in the following example the last `computing c2, a.y is 1` message shouldn't get printed to the console:

```ts
const a = reader({ x: 0, y: 0 });
const c1 = computed(() => {
	console.log("computing c1, a.x is", a.x);
});
const c2 = computed(() => {
	console.log("computing c2, a.y is", a.y);
});

a.x = 1;  // Set _fields bit 0.
a.y = 1;  // Set _fields bit 1.
c1.value; // Console: computing c1, a.x is 1
c2.value; // Console: computing c2, a.y is 1

a.x = 2;  // _fields bits 0 and 1 stay set.
c1.value; // Console: computing c1, a.x is 2
c2.value; // Console: computing c2, a.y is 1
```

One workaround that comes to mind would be to set `a`'s bits back to zero e.g. when the corresponding signal value is read by an effect/computed. But a solution like that may break cases where computeds depends on a overlapping subsets of properties, like in this example (that works now, this is just to illustrate a hypothetical 🙂):

```ts
const a = reader({ x: 0, y: 0, z: 0 });
const c1 = computed(() => {
	console.log("computing c1, a.x + a.y is", a.x + a.y);
});
const c2 = computed(() => {
	console.log("computing c2, a.y + a.z is", a.y + a.z);
});
c1.value; // should recompute
c2.value; // should recompute

a.y = 2;
c1.value; // should recompute
c2.value; // should recompute
```

Issue #103, opened 2022-09-08
question: computed objects, compare function


timglabisch:
when returning objects from a computed function the value always changes.
i guess the compare function here is a simple `===`. And `{} === {}` is `false` .

is there a way to use a more advanced compare function?

i guess the current compare function is here:
(https://github.com/preactjs/signals/blob/d0e5c2d6f1e2e25bc9717099b18d1ea8a16da6f1/packages/core/src/index.ts#L267)

XantreGodlike:
@JoviDeCroock I think this would helpful to provide a way to use custom equality checks. I can provide PR with this changes

XantreGodlike:
@andrewiggins @JoviDeCroock what do you think about custom compare function?

PR #218, opened 2022-09-26
Allow Customizing Diff Logic (via new `shouldUpdate` method)

sghoweri:
This PR provides a way to (optionally) customize how incoming values are evaluated (ie. has this value _actually_ changed) via the new `shouldUpdate` method; a very similar use case to #103 that I also ran into!

Also adds a new test to demonstrate before vs after behavior + a tiny usage example.


XantreGodlike:
Changing prototype is affecting all signals. I thinks it's not a best way to provide custom equality checks

Issue #204, opened 2022-09-23
React Signals not working when renderToStaticMarkup is used in an App


natocTo:
Hi, when I tried a library I found that the signals did not work if in an app is used `renderToStaticMarkup` from `react-dom/server`. It is a bug or expected behavior?

I created an sandbox for test https://codesandbox.io/s/goofy-pascal-6l1dmf

Same apply if `renderToString` is used. Everythings works when I remove an usage of those functions.

developit:
It appears this is only an issue when `renderToStaticMarkup` is used directly from within a component that is currently being rendered by ReactDOM.

natocTo:
Hi, thank you for an example. I will test it once again if I can somehow update my code to something like this.

JoviDeCroock:
This is fixed in 1.2.2 https://codesandbox.io/s/angry-tdd-sy0rr6

PR #219, opened 2022-09-27
React: Replace useReducer with useSyncExternalStore

eddyw:
From Preact Slack discussion 😄 

# Description

`useSyncExternalStore` is the suggested way by React to subscribe to state that lives outside of React which is compatible with their concurrent mode and time-slicing features.

This PR also removes the need to track current owner because hooks run in the context of the current owner so there is no need to really do it. Please check the comments in the code.

### DIFF
--- a/packages/react/src/index.ts
+++ b/packages/react/src/index.ts
@@ -36,17 +36,28 @@ const ProxyHandlers = {
 	 * With the native Proxy, all other calls such as access/setting to/of properties will
 	 * be forwarded to the target Component, so we don't need to copy the Component's
 	 * own or inherited properties.
+	 *
+	 * @see https://github.com/facebook/react/blob/2d80a0cd690bb5650b6c8a6c079a87b5dc42bd15/packages/react-reconciler/src/ReactFiberHooks.old.js#L460
 	 */
 	apply(Component: FunctionComponent, thisArg: any, argumentsList: any) {
 		const store = useMemo(createEffectStore, Empty);
 
 		useSyncExternalStore(store.subscribe, store.getSnapshot, store.getSnapshot);
 
-		const ends = store.updater._start();
-		const kids = Component.apply(thisArg, argumentsList);
-		ends();
+		const stop = store.updater._start();
 
-		return kids;
+		try {
+			const children = Component.apply(thisArg, argumentsList);
+			return children;
+		} catch (e) {
+			// Re-throwing promises that'll be handled by suspense
+			// or an actual error.
+			throw e;
+		} finally {
+			// Stop effects in either case before return or throw,
+			// Otherwise the effect will leak.
+			stop();
+		}
 	},
 };
 
@@ -108,10 +119,8 @@ function createEffectStore() {
 		updater = this;
 	});
 	updater._callback = function () {
-		if (!onChangeNotifyReact) return void unsubscribe();
-
 		version = (version + 1) | 0;
-		onChangeNotifyReact!();
+		if (onChangeNotifyReact) onChangeNotifyReact();
 	};
 
 	return {
@@ -121,10 +130,18 @@ function createEffectStore() {
 
 			return function () {
 				/**
-				 * @todo - we may want to unsubscribe here instead?
-				 * Components wrapped in `memo` no longer get updated if we unsubscribe here.
+				 * Rotate to next version when unsubscribing to ensure that components are re-run
+				 * when subscribing again.
+				 *
+				 * In StrictMode, 'memo'-ed components seem to keep a stale snapshot version, so
+				 * don't re-run after subscribing again if the version is the same as last time.
+				 *
+				 * Because we unsubscribe from the effect, the version may not change. We simply
+				 * set a new initial version in case of stale snapshots here.
 				 */
+				version = (version + 1) | 0;
 				onChangeNotifyReact = undefined;
+				unsubscribe();
 			};
 		},
 		getSnapshot() {
@@ -142,7 +159,8 @@ function WrapJsx<T>(jsx: T): T {
 		}
 
 		if (type && typeof type === "object" && type.$$typeof === ReactMemoType) {
-			return jsx.call(jsx, ProxyFunctionalComponent(type.type), props, ...rest);
+			type.type = ProxyFunctionalComponent(type.type);
+			return jsx.call(jsx, type, props, ...rest);
 		}
 
 		if (typeof type === "string" && props) {
--- a/packages/react/src/index.ts
+++ b/packages/react/src/index.ts
@@ -19,6 +19,8 @@ import { Effect, ReactDispatcher } from "./internal";
 
 export { signal, computed, batch, effect, Signal, type ReadonlySignal };
 
+const Empty = Object.freeze([]);
+
 /**
  * Install a middleware into React.createElement to replace any Signals in props with their value.
  * @todo this likely needs to be duplicated for jsx()...
@@ -55,7 +57,6 @@ function createPropUpdater(props: any, prop: string, signal: Signal) {
 */
 
 let finishUpdate: (() => void) | undefined;
-const updaterForComponent = new WeakMap<() => void, Effect>();
 
 function setCurrentUpdater(updater?: Effect) {
 	// end tracking for the current update:
@@ -64,13 +65,39 @@ function setCurrentUpdater(updater?: Effect) {
 	finishUpdate = updater && updater._start();
 }
 
-function createUpdater(update: () => void) {
+/**
+ * A redux-like store whose store value is a positive 32bit integer (a 'version') to be used with useSyncExternalStore API.
+ * React (current owner) subscribes to this store and gets a snapshot of the current 'version'.
+ * Whenever the 'version' changes, we tell React it's time to update the component (call 'onStoreChange').
+ *
+ * How we achieve this is by creating a binding with an 'effect', when the `effect._callback' is called,
+ * we update our store version and tell React to re-render the component ([1] We don't really care when/how React does it).
+ *
+ * [1]
+ * @see https://reactjs.org/docs/hooks-reference.html#usesyncexternalstore
+ * @see https://github.com/reactwg/react-18/discussions/86
+ */
+function createEffectStore() {
 	let updater!: Effect;
+	let version = 0;
+
 	effect(function (this: Effect) {
 		updater = this;
 	});
-	updater._callback = update;
-	return updater;
+
+	return {
+		updater,
+		subscribe(onStoreChange: () => void) {
+			updater._callback = function () {
+				version = (version + 1) & ~(1 << 31);
+				onStoreChange();
+			};
+			return function _unsubscribe() {};
+		},
+		getSnapshot() {
+			return version;
+		},
+	};
 }
 
 /**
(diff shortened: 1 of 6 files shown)


packages/react/src/index.ts
```diff
@@ -64,13 +65,39 @@ function setCurrentUpdater(updater?: Effect) {
...
+ * A redux-like store whose store value is a positive 32bit integer (a 'version') to be used with useSyncExternalStore API.
+ * React (current owner) subscribes to this store and gets a snapshot of the current 'version'.
+ * Whenever the 'version' changes, we tell React it's time to update the component (call 'onStoreChange').
+ *
+ * How we achieve this is by creating a binding with an 'effect', when the `effect._callback' is called,
+ * we update our store version and tell React to re-render the component ([1] We don't really care when/how React does it).
+ *
+ * [1]
+ * @see https://reactjs.org/docs/hooks-reference.html#usesyncexternalstore
+ * @see https://github.com/reactwg/react-18/discussions/86
```
eddyw, line 78: Maybe this one is more relevant 😅 
```suggestion
 * @see https://github.com/reactjs/rfcs/blob/main/text/0214-use-sync-external-store.md
```

marvinhagemeister (approved):
This is a great idea to use `useSyncexternalStore` instead of `useReducer` 👍

eddyw:
@marvinhagemeister updated PR to use `use-sync-external-store/shim`
Added as peer dep, it should be a peer dep, right? 😅

marvinhagemeister:
@eddyw Let's use a normal dependency for that to keep the packages to a minimum that the user has to install

eddyw:
@marvinhagemeister makes sense, done!

packages/react/src/index.ts
```diff
@@ -64,13 +66,39 @@ function setCurrentUpdater(updater?: Effect) {
...
-	return updater;
+
+	return {
+		updater,
+		subscribe(onStoreChange: () => void) {
+			updater._callback = function () {
+				version = (version + 1) & ~(1 << 31);
+				onStoreChange();
+			};
+			return function _unsubscribe() {};
```
marvinhagemeister, line 96: We should probably clean up the `effect()` here. Every `effect()` call returns an unsubscribe functions so we can return that one here.

```js
const unsubscribe = effect(() => ...);
```

eddyw: I did so at first, but cleaning up the effect makes the test `should consistently rerender in strict mode` fail.
The previous implementation didn't unsubscribe either so I thought maybe it was intentional 🤔

packages/react/src/index.ts
```diff
@@ -19,6 +20,8 @@ import { Effect, ReactDispatcher } from "./internal";
 
 export { signal, computed, batch, effect, Signal, type ReadonlySignal };
 
+const Empty = Object.freeze([]);
```
developit, line 23: No need to freeze here - we could enforce readonly via TS, but tbh I think it's fine to just call it EMPTY and leave it mutable ([we do this in Preact core](https://github.com/preactjs/preact/blob/dfd45aaa02c9812569aca29dd52afae2a9d57016/src/constants.js#L2) too).

```suggestion
const Empty = [];
```

packages/react/src/index.ts
```diff
@@ -64,13 +66,39 @@ function setCurrentUpdater(updater?: Effect) {
...
 		updater = this;
 	});
-	updater._callback = update;
-	return updater;
+
+	return {
+		updater,
+		subscribe(onStoreChange: () => void) {
+			updater._callback = function () {
+				version = (version + 1) & ~(1 << 31);
```
developit, line 93: This might be a bit smaller + cheaper and should overflow to negative integers to avoid collisions similarly to what you had:

```suggestion
				version = (version + 1) | 0;
```

packages/react/src/index.ts
developit: Is it possible the useMemo() is running twice in strict mode? I'm not up-to-date on the funky semantics of React's modes, but afaik that would cause unsubscription and maybe wouldn't re-subscribe properly?

developit:
I think we might be missing a bunch of tests for the React adapter - I'd like to add those first before we merge this just to make sure it isn't changing the semantics (or rather that it's only improving them). We should be able to copy over some of the test cases from the Preact adapter.

packages/react/src/index.ts
eddyw: @developit yes, please check this commit: https://github.com/preactjs/signals/pull/219/commits/2e9ea59ce63ec10be84703b31a88e7491cc4cb7e

I'm not sure if this'd be the best approach though. Another way could be to rely on the React current owner internal (which is set only on dev) but I tried to make it rely as less as possible on the internals.

I left comments in the code. I tried my best to explain it 😅 
In prod, it'll always unsubscribe, on dev it lazily unsubscribes itself. This can be tested by setting `__DEV__ = false` in which case the StrictMode is the only test that'll fail.

I'm also not quite sure how you'd test using the React prod entry point, apparently if tests are executed with `NODE_ENV=production` this `act` thing errors with kinda "_isn't supported with production build_". Do these Netlify demos use the production one?

packages/react/src/index.ts
```diff
@@ -19,6 +20,13 @@ import { Effect, ReactDispatcher } from "./internal";
 
 export { signal, computed, batch, effect, Signal, type ReadonlySignal };
 
+const Empty = [] as const;
+
+/**
+ * React uses a different entry-point depending on NODE_ENV env var
+ */
+const __DEV__ = process.env.NODE_ENV !== "production";
```
developit, line 28: We'll need to find a way to ignore the double effect thing without relying on this - it breaks in ESM.

packages/react/src/index.ts
```diff
@@ -97,30 +153,31 @@ Object.defineProperties(Signal.prototype, {
...
+			// context changes.
 			lock = true;
-			const rerender = api.useReducer(UPDATE, {})[1];
+
+			const store = api.useMemo(createEffectStore, Empty);
+
+			useSyncExternalStore(
+				store.subscribe,
+				store.getSnapshot,
+				store.getSnapshot
```
developit, line 175: I believe the third argument is optional and defaults to the 2nd here?

```suggestion
				store.getSnapshot
```

eddyw:
@developit I did some testing yesterday with NextJS (dev & prod) builds locally and it doesn't work reliably. It did work in a CRA app though (only checked dev on StackBlitz).

I have followed a different approach locally, since we do wrap the `React.createElement`, I did something like this:
```ts
function WrapCreateElementFactory<T>(maybeFunction: T): T {
	return function (type: any, props: any, ...rest: any[]) {
		if (typeof type === "function" && !(type instanceof Component)) {
			type = ProxyFunctionalComponent(type); // <<<< this
		} else if (
			typeof type === "object" &&
			type != null &&
			type.$$typeof === MemoElementType
		) {
			type.type = ProxyFunctionalComponent(type.type); // <<<< this for memo
		}
		return maybeFunction.call(maybeFunction, type, props, ...rest);
	} as any as T;
}

React.createElement = WrapCreateElementFactory(CreateElement);
// JsxDev.jsx && (JsxDev.jsx = WrapCreateElementFactory(JsxDev.jsx));
// (...) also wraps jsxs and jsxDEV for either react/jsx-runtime or react/jsx-dev-runtime
```

The `ProxyFunctionalComponent` just wraps the component in a `new Proxy(Component, ProxyHandlers)` (also have fallback if Proxy isn't supported) and uses a `apply` proxy trap:
```ts
const ProxyHandlers = {
	apply(Component: FunctionComponent, thisArg: any, argumentsList: any) {
		const EffectStore = useMemo(createEffectStore, Empty);

		useSyncExternalStore(
			EffectStore.subscribe,
			EffectStore.getSnapshot,
			EffectStore.getSnapshot
		);

		const EndEffect = EffectStore.updater._start();
		const Children = Component.apply(thisArg, argumentsList);
		EndEffect();

		return Children;
	},
};
```

So basically, this way we don't even need to hook into the current dispatcher but just let React call the function (internally React does something like `children = Component()`) and we can run hooks normally without worrying which ones are invalid (there are like ~4 kinds of dispatchers in dev).

What do you think about this? My local changes using this approach pass all tests and it works on NextJS dev & prod builds. NextJS uses jsx-runtime / jsx-dev-runtime so I wrote this `WrapCreateElementFactory` function which works the same for `React.createElement`.

Do you want to try this approach instead? If so I can commit my local changes 😅.
Otherwise I'm not sure right now how to achieve this reliably for prod vs dev because it's hard to know if React did call the function component or not when setting the dispatcher which makes it work unreliably on client vs server rendering or prod vs dev.

Alternatively, _maybe_ it'd make sense to provide a `React.memo`-like wrapper? Maybe a way to opt-in into having all components wrapped?

packages/react/src/index.ts
eddyw: I don't remember if it warns when it fallbacks to the 1st one 🤔 
I'll check that and remove it if it's not needed.

packages/react/src/index.ts
```diff
@@ -15,64 +17,165 @@ import {
...
+function WrapJsx<T>(jsx: T): T {
+	if (typeof jsx !== "function") return jsx;
+
+	return function (type: any, props: any, ...rest: any[]) {
+		if (typeof type === "function" && !(type instanceof Component)) {
+			return jsx.call(jsx, ProxyFunctionalComponent(type), props, ...rest);
+		}
+
+		if (type && typeof type === "object" && type.$$typeof === ReactMemoType) {
+			return jsx.call(jsx, ProxyFunctionalComponent(type.type), props, ...rest);
```
eddyw, line 145: This should set `type.type = ProxyFunctionalComponent(...)` and not unwrap the memo component 🤦

packages/react/test/index.test.tsx
```diff
@@ -179,5 +200,26 @@ describe("@preact/signals-react", () => {
...
+			for (let i = 0; i < 3; i++) {
+				act(() => {
+					count.value += 1;
+					render(<Test />);
+				});
+				expect(scratch.textContent).to.equal(
+					`<code>${count.value}</code><code>${count.value}</code>`
+				);
+			}
+		});
```
eddyw, line 223: This resolves issue #204

packages/react/src/index.ts
```diff
@@ -15,64 +17,183 @@ import {
...
+	 * Omitting this for now.
+	 *
+	 * @example - works with Proxy, doesn't with wrapped function.
+	 * ```
+	 * const el = <SomeFunctionalComponent />
+	 * el.type.someOwnOrInheritedProperty;
+	 * el.type.defaultProps;
+	 * ```
+	 */
+	const WrappedComponent = function () {
```
marvinhagemeister, line 91: Really appreciate that you went the extra mile and added support for the non-proxy browsers too 👍

marvinhagemeister (approved):
This is fantastic PR! Really like the way you documented the why's of the changes in code comments. Thank you so much for the PR ❤️

packages/react/src/index.ts
```diff
@@ -2,11 +2,13 @@ import {
 	useRef,
 	useMemo,
 	useEffect,
-	// @ts-ignore-next-line
-	// eslint-disable-next-line @typescript-eslint/no-unused-vars
-	__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED as internals,
```
wojtekmaj, line 7: This is huge!

ADTC: Before we get too excited, let's remember that it reappeared in https://github.com/preactjs/signals/blob/main/packages/react/runtime/src/auto.ts

PR #222, opened 2022-09-28
Improvment: Replaced the mv command with move-cli

mohammadraufzahed:
Replaced the mv command with move-cli to make it work in any os that can run nodejs.


rschristian:
Superseded by #555

PR #232, opened 2022-10-06
Simplify effect change checking (and make effect cycle detection more accurate as a side-effect)

jviide:
This pull request simplifies effect change checking by removing the OUTDATED flag usage from effects. The flag itself can't be completely removed from the codebase, as computed signals still use it. Still, there are some delicious byte size savings gained from this change.

The OUTDATED flag became unnecessary after the changes introduced in #205.

As a nice side-effect the effect cycle detection is now more accurate. There is a new test case for that. The test case does collide with #231, and should be removed if/when that pull request is merged. The code simplifications are still valid regardless.

### DIFF
--- a/.changeset/pretty-students-look.md
+++ b/.changeset/pretty-students-look.md
@@ -1,0 +1,5 @@
+---
+"@preact/signals-core": patch
+---
+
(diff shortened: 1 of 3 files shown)


marvinhagemeister (approved):
Nice!! Very good catch on spotting those size savings 💯

PR #243, opened 2022-10-15
Use WeakMap to hold proxy map in react

melnikov-s:
A Regular `Map` is being used to hold a map of proxied components , this can lead to unbounded memory leaks if for example a component is created within another.

ie:

```
function OutterComponent(props) {
   function InnerComponent(props) {...}

   return <InnerComponent />
}

```

Leads to a new entry in the map on each render of `OutterComponent` with no way to collect. `WeakMap` should do the trick here.

### DIFF
--- a/packages/react/src/index.ts
+++ b/packages/react/src/index.ts
@@ -24,7 +24,7 @@ export { signal, computed, batch, effect, Signal, type ReadonlySignal };
 const Empty = [] as const;
 const ReactElemType = Symbol.for("react.element"); // https://github.com/facebook/react/blob/346c7d4c43a0717302d446da9e7423a8e28d8996/packages/shared/ReactSymbols.js#L15
 const ReactMemoType = Symbol.for("react.memo"); // https://github.com/facebook/react/blob/346c7d4c43a0717302d446da9e7423a8e28d8996/packages/shared/ReactSymbols.js#L30
-const ProxyInstance = new Map<FunctionComponent<any>, FunctionComponent<any>>();
+const ProxyInstance = new WeakMap<FunctionComponent<any>, FunctionComponent<any>>();
 const SupportsProxy = typeof Proxy === "function";
 
 const ProxyHandlers = {
--- a/.changeset/breezy-ants-trade.md
+++ b/.changeset/breezy-ants-trade.md
@@ -1,0 +1,5 @@
+---
+"@preact/signals-react": patch
+---
+
+Replace `Map` useage with `WeakMap`


marvinhagemeister:
Thanks for making a PR. We're there more use cases you were thinking of where switching to `WeakMap` would help? The scenario where a component is created _inside_ another component is heavily discouraged. We've treated that as a bug in the code as it's catastrophic for performance even without signals, because it will always force that subtree to unmount and remount every time.

melnikov-s:
Yeah it's not a common or good pattern but I have seen it done, these components might be shallow or memo'd as well so less catastrophic. Component factories would also have this problem or hooks that return (memoized) components to render as part of their api. 

It's hard to predict how code is written in the wild and since `createElement` is being patched and will effect every component and not just those using signals I don't see a downside of being safe and using `WeakMap` here. 

As an aside, the approach of using a proxied component is neat but:

```
function Component() {}
const element = React.CreateElement(Component);
element.type === Component; // false if you import `@preact/signals-react`  true otherwise
```

That's going to break testing libraries that query on the Component. eg: enzyme

packages/react/src/index.ts
```diff
@@ -24,7 +24,7 @@ export { signal, computed, batch, effect, Signal, type ReadonlySignal };
 const Empty = [] as const;
 const ReactElemType = Symbol.for("react.element"); // https://github.com/facebook/react/blob/346c7d4c43a0717302d446da9e7423a8e28d8996/packages/shared/ReactSymbols.js#L15
 const ReactMemoType = Symbol.for("react.memo"); // https://github.com/facebook/react/blob/346c7d4c43a0717302d446da9e7423a8e28d8996/packages/shared/ReactSymbols.js#L30
-const ProxyInstance = new Map<FunctionComponent<any>, FunctionComponent<any>>();
+const ProxyInstance = new WeakMap<FunctionComponent<any>, FunctionComponent<any>>();
```
eddyw, line 27: There are some runtimes that do not have WeakMap but do have Map.
Maybe we could do something like:
```ts
const ProxyInstance = new (typeof WeakMap === "function"
  ? WeakMap
  : Map
)();
```

marvinhagemeister: I think using `WeakMaps` is fine. They are supported even in IE11: https://caniuse.com/?search=JavaScript%20WeakMap

marvinhagemeister:
> It's hard to predict how code is written in the wild and since `createElement` is being patched and will effect every component and not just those using signals I don't see a downside of being safe and using `WeakMap` here. 

Good point about `createElement`. Didn't think of that as the Preact adapter is more fresh on my mind. There we don't need to do such things.

marvinhagemeister (approved):

Issue #241, opened 2022-10-15
Errors from dependent effects are propagated to value setter


felixschorer:
Given a signal `foo` and an effect which throws an error whenever `foo` holds the value `true`.  

When the effect is created when `foo` already holds the value `true`,  
then an error will be thrown upon creating that effect.
```ts
const foo = signal(true);

// throws
effect(() => {
  if (foo.value) throw new Error("test");
});
```

However, when the effect is created while `foo` holds the value `false`,  
then an error will be thrown upon setting the value of `foo` to `true`.
```ts
const foo = signal(false);

effect(() => {
  if (foo.value) throw new Error("test");
});

// throws
foo.value = true;
```
This behavior is unexpected.
The presence of a faulty effect may break code, which otherwise works fine in isolation.
The errors caused by an effect can only appropriately be handled by the effect itself or logged.

A fix could be to re-throw those errors inside a new micro task.
- Effects could still run eagerly.
- Errors can be handled by a global error handler to simply log them.
- The code, which is setting the value, wouldn't break depending if there is a faulty effect present or not.

----

Another (probably related) issue is that effects aren't (and can't be) disposed properly, when the initial execution throws an error.

```ts
const foo = signal(true);

try {
  const dispose = effect(() => {
    if (foo.value) {
      throw new Error("test");
    } else {
      console.log(foo.value);
    }
  });
} catch (err) { /* ignore */ }

// Logs `true` to the console, i.e., the effect is still subscribed to `foo`.
// The effect also cannot be disposed as `dispose` is out of scope.
foo.value = false;
```

marvinhagemeister:
> This behavior is unexpected.

I'd love to dive deeper on this on how you reached that conclusion. In both examples provided I'd expect an error to be thrown.

The last example with being unable to dispose them seems like a bug that we should address regardless.

developit:
Might make sense to mark this issue as fixed? #245 addresses the idiosyncrasy between initial and update error propagation, which is the "unexpected" part.

Direct error propagation to the setter is an explicit design choice we made in order to avoid Signals inventing a new error propagation/handling mechanism where one already exists. The error propagation/handling logic in Signals isn't supposed to allow applications to continue to function partially in an error state, it's to prevent the graph from being incorrect such that applications with error recovery logic can function in a recovered state.

marvinhagemeister:
Good point, marking as resolved by #245

felixschorer:
[...]

PR #245, opened 2022-10-17
Fix effect behavior when first run throws

jviide:
This pull request fixes a bug mentioned by @felixschorer in #241. If the first run of an effect throws the effect should be disposed.


marvinhagemeister (approved):
Awesome 👍

PR #262, opened 2022-11-04
Fix typo

elliotwaite:


elliotwaite:
I noticed additional typos while reading over the tests, so I decided to just close this PR and create a separate one that included those fixes as well: https://github.com/preactjs/signals/pull/263

PR #263, opened 2022-11-04
Fix typos and other small changes

elliotwaite:
Mainly typo fixes, but also made a change to [this test](https://github.com/preactjs/signals/blob/main/packages/core/test/signal.test.tsx#L752). I noticed that setting `s.value = 2;` after `e._dispose();` wasn't actually testing if the dispose function worked because the value of `s` was already 2 at that point. I verified this by commenting out the `e._dispose();` line and the tests still passed. So I changed that line to `s.value = 3;` which now makes the test only pass if `e._dispose();` is called.

Also, I edited a part of [one of the test diagrams](https://github.com/preactjs/signals/blob/main/packages/core/test/signal.test.tsx#L1457) from `D <- we don't listen to C` to `D <- we don't listen to D`, but I wasn't sure if that diagram should actually be changed more. I noticed that the diagram was probably copied from the test above it and maybe that note of `we don't listen to D` should actually be removed because we actually do listen to `D` in parts of that test. Or maybe it could be changed to `D <- the event never listens to C`, since it's the event that tracks the changes we are testing. But I wasn't sure the best way to change it other than just fixing the `C` to `D` typo. If you want me to update this PR with an alternative correction for that diagram, let me know.

### DIFF
--- a/packages/core/test/signal.test.tsx
+++ b/packages/core/test/signal.test.tsx
@@ -129,7 +129,7 @@ describe("signal", () => {
 			expect(spy).not.to.be.called;
 		});
 
-		it("should not cause surrounding effect subscribe to changes to a signal accessed in the callback", () => {
+		it("should not cause surrounding effect to subscribe to changes to a signal accessed in the callback", () => {
 			const spy = sinon.spy();
 			const a = signal(0);
 			const b = signal(0);
@@ -663,7 +663,7 @@ describe("effect()", () => {
 			expect(newSpy).to.be.called;
 		});
 
-		it("should returns a function for closing the effect scope from _start", () => {
+		it("should return a function for closing the effect scope from _start", () => {
 			const s = signal(0);
 
 			let e: any;
@@ -749,7 +749,7 @@ describe("effect()", () => {
 			spy.resetHistory();
 
 			e._dispose();
-			s.value = 2;
+			s.value = 3;
 			expect(spy).not.to.be.called;
 		});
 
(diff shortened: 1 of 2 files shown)


packages/core/test/signal.test.tsx
```diff
@@ -1449,12 +1449,12 @@ describe("computed()", () => {
...
 
 		it("should only subscribe to signals listened to", () => {
-			// Here both "B" and "C" are active in the beginnning, but
+			// Here both "B" and "C" are active in the beginning, but
 			// "B" becomes inactive later. At that point it should
 			// not receive any updates anymore.
 			//    *A
 			//   /   \
-			// *B     D <- we don't listen to C
+			// *B     D <- we don't listen to D
```
eddyw, line 1457: I think this meant that `D` doesn't listen to changes in `C` because ..

> Here both "B" and "C" are active in the beginning

I think this means "active" as the effect is listening to `C` (so `B` is a target in `A` and `C` is a target in `B`) but then it calls `unsub` to unsubscribe which makes `C` contain to targets, so it unsubscribes from `B` and in the same way, `B` will have no targets, so it'll unsubscribe from `A`.

Then setting the value of a `a.value = "aa"` is testing that `B` and `C` compute functions aren't called but `D` must update (`D` doesn't listen to neither `C` nor `B`)

elliotwaite: Ah okay, I reverted that change.

I thought that part of the diagram was just left over from it being copied from the test above it and the author forgot to update that part of it. It still seems strangely worded to me. I usually think of listening as only going up the graph. Something like "C no longer listens to A, but D does" would make more sense to me. But I'm okay with just leaving it as is if you think how it is currently written was intentionally by the author.

rschristian (approved):
Thanks!

PR #266, opened 2022-11-04
Proposal: use watcher

### LINKED ISSUE #247: Sync between useSignal and its parameter
For both React and Preact, the `useSignal` hook implementation is currently defined as:

https://github.com/preactjs/signals/blob/ccb677bdb26ef836d3fac39d30d8f6371693e60f/packages/preact/src/index.ts#L334-L336

This implementation would cause the returned signal's `.value` to only be kept in sync with the `value` parameter on the very first call.

```js
let param = 0;

// inside a compoenent

const num = useSignal(param); // num.value is 0

// re-render component with a modified param

param = 5;

const num = useSignal(param); // num.value is still 0
```

One way to get around this behavior is to immediately set the `.value` on the returned signal:

https://github.com/preactjs/signals/blob/ccb677bdb26ef836d3fac39d30d8f6371693e60f/packages/preact/src/index.ts#L70-L71

I am wondering, is there a specific reason that the `.value` is not set directly in the `useSignal` function? One possible limitation I can see is that batching multiple signal updates would become not possible.

I do think however that the benefit of synchronizing between a signal's `.value` and its `value` parameter outweighs that limitation.

billybimbob:
As discussed in #247, this PR introduces a new hook, `useWatcher`, which converts a non-signal value into a signal.

### Motivation

Signals are a powerful primitive, but one main restriction of them is that signal dependencies can only be established between other signals. This means that deriving values between a non-signal and a signal can potentially not update correctly:

```ts
function Foo({ s, n }: { s: ReadonlySignal<number>, n: number }) {

    const combined = useComputed(() => s.value * n * n);
    // combined only recomputes if s is modified

    return <p>{combined}</p>;
}
```

Specific scenarios where this limitation is apparent is when using third-party libraries that do not store state with signals. The `useWatcher` hook is a solution to this restriction, as it essentially wraps a non-signal value in a signal. This as a result allows for changes to the non-signal value to be seen by other signals:

```ts
function Foo({ s, n }: { s: ReadonlySignal<number>, n: number }) {
    const $n = useWatcher(n);

    const combined = useComputed(() => s.value * $n.value * $n.value);
    // combined now updates correctly

    return <p>{combined}</p>;
}
```

While this hook could just be created by end-users, the most optimal implementation of the `useWatcher` breaks typical React convention, which can lead to pitfalls for said end-users.

### Implementation

While all current testcases are passing, one area of uncertainty is with the Preact adapter. Setting the signal `.value` in `useWatcher` may cause extra Component updates to trigger when they shouldn't, so some verification on that end would be appreciated.

Resolves #247

### DIFF
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -353,6 +353,12 @@ export function useSignalEffect(cb: () => void | (() => void)) {
 	}, []);
 }
 
+export function useWatcher<T>(value: T) {
+	const watcher = useSignal(value);
+	watcher.value = value;
+	return watcher as ReadonlySignal<T>;
+}
+
 /**
  * @todo Determine which Reactive implementation we'll be using.
  * @internal
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -54,7 +54,7 @@ function createUpdater(update: () => void) {
 // 	if (typeof value !== "object" || value == null) return false;
 // 	if (value instanceof Signal) return true;
 // 	// @TODO: uncomment this when we land Reactive (ideally behind a brand check)
-// 	// for (let i in value) if (value[i] instanceof Signal) return true;
+// 	// for (const i in value) if (value[i] instanceof Signal) return true;
 // 	return false;
 // }
 
@@ -67,10 +67,9 @@ function Text(this: AugmentedComponent, { data }: { data: Signal }) {
 
 	// Store the props.data signal in another signal so that
 	// passing a new signal reference re-runs the text computed:
-	const currentSignal = useSignal(data);
-	currentSignal.value = data;
+	const currentSignal = useWatcher(data);
 
-	const s = useMemo(() => {
+	const $s = useMemo(() => {
 		// mark the parent component as having computeds so it gets optimized
 		let v = this.__v;
 		while ((v = v.__!)) {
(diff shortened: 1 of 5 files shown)


ismi29ch (approved):

packages/preact/src/index.ts
```diff
@@ -351,6 +352,12 @@ export function useSignalEffect(cb: () => void | (() => void)) {
 	}, []);
 }
 
+export function useWatcher<T>(value: T) {
+	const watcher = useSignal(value);
+	watcher.value = value;
+	return watcher as ReadonlySignal<T>;
+}
```
jamesarosen, line 359: With this implementation, I sometimes get the following warning:

> react-dom.development.js:86 Warning: Cannot update a component (`Foo`) while rendering a different component (`Bar`). To locate the bad setState() call inside `Bar`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render

I changed it to this and the warning went away:
```ts
export default function useWatcher<T>(value: T): ReadonlySignal<T> {
  const watcher = useMemo<Signal<T>>(() => {
    return signal(value)
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [])

  useEffect(() => {
    watcher.value = value
  }, [value, watcher])
}
```

It's absolutely possible that the fault is entirely within my application and the change above papers over a real problem. It's also possible that setting `watcher.value = value` outside of a React hook is causing problems with React. I haven't narrowed that down yet.

XantreGodlike: I think we should ignore it, its not something bad

XantreGodlike:
Btw, i don't thinks its major release

XantreGodlike:
I think it shouldn't be in adapter lib, it will have good fit inside of utility libs. I will create collection of hooks for signals

packages/preact/src/index.ts
XantreGodlike: I know how to fix it: wrap component in batch, so it will not try to rerender instantly. What do you think? @andrewiggins

PR #268, opened 2022-11-06
Update Preact VNode adapter

### LINKED ISSUE #256: SSR does not render text bindings
Upgrading to the recent versions of [preact-render-to-string](https://github.com/preactjs/preact-render-to-string) seems to cause signals passed as text bindings to render blank instead of the expected signal value.

This behavior can be seen when running the testcases with a modified `preact-render-to-string` dependency version:

```jsonc
// packages/preact/package.json

"devDependencies": {
	"preact": "10.9.0",
	"preact-render-to-string": "^5.2.5"
}
```

Edit: [Comparing the changes](https://github.com/preactjs/preact-render-to-string/compare/60075a5a7389d638d535c85f3706739e9ba932bc...v5.2.5) to `preact-render-to-string`, I believe that the cause for the text binding regression is [this change](https://github.com/preactjs/preact-render-to-string/blob/359a58e8f1f6c501f7087710335d2addba818d90/src/index.js#L214-L215).

From how I understand the signal Text node [implementation](https://github.com/preactjs/signals/blob/main/packages/preact/src/index.ts#L65), the `Text` component does have a defined constructor, as a result of `Text` extending from `Signal`.

billybimbob:
Preact expects VNodes to have an undefined constructor. This PR updates the Preact adapter so that signal objects match this expectation. The undefined constructor will only affect signal **objects** while the Signal Prototype is unaffected.

Fixes #256

### DIFF
--- a/packages/preact/test/index.test.tsx
+++ b/packages/preact/test/index.test.tsx
@@ -1,4 +1,4 @@
-import { signal, useComputed } from "@preact/signals";
+import { signal, computed, useComputed, Signal } from "@preact/signals";
 import { createElement, render } from "preact";
 import { setupRerender, act } from "preact/test-utils";
 
@@ -15,6 +15,16 @@ describe("@preact/signals", () => {
 
 	afterEach(() => {
 		render(null, scratch);
+	});
+
+	describe("inheritance", () => {
+		it("should have signals inherit from Signal", () => {
+			expect(signal(0)).to.be.instanceof(Signal);
+		});
+
+		it("should have computed inherit from Signal", () => {
+			expect(computed(() => 0)).to.be.instanceof(Signal);
+		});
 	});
 
 	describe("Text bindings", () => {
--- a/packages/preact/package.json
+++ b/packages/preact/package.json
@@ -45,6 +45,6 @@
 	},
 	"devDependencies": {
 		"preact": "10.9.0",
-		"preact-render-to-string": "^5.2.4"
+		"preact-render-to-string": "^5.2.5"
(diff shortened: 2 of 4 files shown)


JoviDeCroock (approved):

packages/preact/test/index.test.tsx
```diff
@@ -17,6 +17,16 @@ describe("@preact/signals", () => {
 		render(null, scratch);
 	});
 
+	describe("inheritance", () => {
```
JoviDeCroock, line 20: As we have `preact-render-to-string` in our deps, should we add a test for `<p>{signal}</p>`?

billybimbob: Would the test that you're suggesting be different than the SSR tests in [`packages/preact/test/ssr.test.tsx`](https://github.com/preactjs/signals/blob/55d5becbb41c2faec62b9588937f8c74f0d65152/packages/preact/test/ssr.test.tsx#L9)?

JoviDeCroock: Aha, updating initially broke the tests?

billybimbob: Yeah, I can't remember which ones specifically, but iirc, 7 of the text binding tests were failing.

The inheritance tests also might be redundant, since duplicate tests are ran in `packages/core/test/signal.test.tsx`. My thought process for adding those tests was that modifying the Signal prototype may potentially mess with `instanceof`.

Though thinking about it now, it might be better to just remove those tests.

PR #271, opened 2022-11-09
Type Signal as valid React JSX

### LINKED ISSUE #261: Typescript error when passing signal directly into JSX
When passing a signal directly into JSX in an attempt to bypass VDOM leads to the below TS error. It does work as expected from a functional standpoint however.

`Type 'ReadonlySignal<string>' is not assignable to type 'ReactNode'.ts`

Screen Shot 2022-11-03 at 10 13 50 PM

`"@preact/signals-react": "^1.2.1"`
`"@types/react": "^18.0.24"`

Just wondering what the recommended approach is around this? Thank you!

billybimbob:
This PR updates the typing of the React signal adapter to enable the text binding. General refactors to the React adapter typing are also included.

The reason for this change is because the implementation of the React signal adapter [supports direct text binding](https://github.com/preactjs/signals/blob/55d5becbb41c2faec62b9588937f8c74f0d65152/packages/react/test/index.test.tsx#L34-L40).

A couple `@ts-ignore` were also added to the core package to account for how Typescript interprets ES5 classes.

Fixes #261

### DIFF
--- a/.changeset/metal-emus-design.md
+++ b/.changeset/metal-emus-design.md
@@ -1,0 +1,5 @@
+---
+"@preact/signals-react": patch
+---
+
+type Signal as a React Element
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -157,11 +157,13 @@ function addDependency(signal: Signal): Node | undefined {
 	return undefined;
 }
 
+// @ts-ignore internal Signal is viewed as a function
 declare class Signal<T = any> {
 	/** @internal */
 	_value: unknown;
 
-	/** @internal
+	/**
+	 * @internal
 	 * Version numbers should always be >= 0, because the special value -1 is used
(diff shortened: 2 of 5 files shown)


MrHBS:
@billybimbob Isn’t this a Preact only optimization?

billybimbob:
> @billybimbob Isn’t this a Preact only optimization?

Both the implementation and the testcases for the React adapter seem to indicate that direct text binding is actually possible. #261 also confirms that the text binding does functionally work.

The main current issue seems to be that the React JSX prevents signals as text nodes.

Donneto:
Any idea for when this might be approved/implemented? Thanks for the hard work. Much appreciated.

jll90:
I just ran into the same issue and I'm wondering if this could be merged soon. Thanks!

JoviDeCroock (approved):

PR #277, opened 2022-11-29
Updated Readme Link to Valid Tag

Dhananjay-JSR:
Computed and Effect Anchor Tags were Linked To incorrect reference


rschristian (approved):
Good catch, thanks!

PR #278, opened 2022-12-04
add description to package.json

nichoth:
Add the description to `package.json`, so something shows up when I look at dependencies on github.

Screen Shot 2022-12-04 at 1 42 06 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,8 +1,5 @@
 {
 	"name": "preact-signals",
-	"version": "1.0.0",
-	"description": "Manage state with style in every framework",
-	"main": "index.js",
 	"private": true,
 	"scripts": {
 		"prebuild": "rimraf packages/core/dist/ packages/preact/dist",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
 	"name": "preact-signals",
 	"version": "1.0.0",
-	"description": "",
+	"description": "Manage state with style in every framework",
 	"main": "index.js",
 	"private": true,
 	"scripts": {
--- a/packages/preact/package.json
+++ b/packages/preact/package.json
@@ -2,7 +2,7 @@
 	"name": "@preact/signals",
 	"version": "1.1.2",
 	"license": "MIT",
-	"description": "",
+	"description": "Manage state with style in every framework",
 	"keywords": [],
 	"authors": [
 		"The Preact Authors (https://github.com/preactjs/signals/contributors)"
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
 	"name": "preact-signals",
 	"version": "1.0.0",
-	"description": "",
+	"description": "Manage state with style in every framework",
```
rschristian, line 4: Setting this will do nothing, as this is the root `package.json` for the monorepo. It's here to help with common tooling and whatnot, it's not a package itself.

You'll instead want to set this in each package, so here in the case of `@preact/signals`:

https://github.com/preactjs/signals/blob/d3a566e1842ab93505070a403f800e3f4a999557/packages/preact/package.json#L1-L5

nichoth: Yes, sorry for the mistake.

rschristian: No need to apologize!

We'd still love your contribution, if you want to update the correct files & re-open this PR I'd be happy to merge it in

package.json
```diff
@@ -1,5 +1,8 @@
 {
 	"name": "preact-signals",
+	"version": "1.0.0",
+	"description": "Manage state with style in every framework",
+	"main": "index.js",
```
rschristian, line 5:
```suggestion
```

Sorry, I removed these in #279 as they're rather useless for a monorepo root. Our mistake for still having them around!

package.json
nichoth: ✅

packages/preact/package.json
```diff
@@ -2,7 +2,7 @@
 	"name": "@preact/signals",
 	"version": "1.1.2",
 	"license": "MIT",
-	"description": "",
+	"description": "Manage state with style in every framework",
```
rschristian, line 5:
```suggestion
	"description": "Manage state with style in Preact",
```

Let's change this to 'in Preact' as this is the Preact-specific package.

packages/react/package.json
```diff
@@ -2,7 +2,7 @@
 	"name": "@preact/signals-react",
 	"version": "1.2.1",
 	"license": "MIT",
-	"description": "",
+	"description": "Manage state with style in every framework",
```
rschristian, line 5:
```suggestion
	"description": "Manage state with style in React",
```

Same thing here

rschristian (approved):

rschristian:
Thanks for catching that these were missing!

Issue #228, opened 2022-10-03
useSignalEffect should execute after changes have been applied to the DOM


luisherranz:
The callback passed to `useSignalEffect` should be executed after the changes (if any) have been applied to the DOM. Otherwise, it won't be a direct replacement for `useEffect`.

For example, people expect this to work, even if `ref` appears when `open` switches from `false` to `true`, because `useEffect` waits until changes have been applied to the DOM:

```js
useEffect(() => {
  if (open) ref.current.focus();
}, [open]);
```

But with the current implementation of `useSignalEffect`, the callback runs before the DOM changes, and therefore `ref.current` may not exist yet:

```js
useSignalEffect(() => {
  if (open.value) ref.current.focus();
});
```

[Full example in StackBlitz](https://stackblitz.com/edit/vitejs-vite-e1qt21?file=src%2Fmain.jsx,index.html).

_Other reactive libraries introduce a `tick` function ([Vue](https://vuejs.org/api/general.html#nexttick), [Svelte](https://svelte.dev/docs#run-time-svelte-tick), [Alpine](https://alpinejs.dev/magics/nextTick)), but I'd prefer if `useSignalEffect` could mimic `useEffect` because the familiarity and ergonomics are better. I.e., access to signals after the `tick()` call are not tracked._

marvinhagemeister:
That's a very good point. I hadn't thought about the interplay with ref's.

JoviDeCroock:
Hmm, that makes sense. I guess the sync escape hatch is possible but leveraging the commit options hook would allow us to immitate the timing of a regular useEffect

luisherranz:
If you give me some instructions, I can work on the PR.

souporserious:
Ran into this as well, I ended up [creating a reactive ref](https://codesandbox.io/s/reactive-callback-ref-l6qhv9) that seems to do the trick and essentially would work like a `tick` mechanism:

```ts
const node = useSignal<HTMLElement | null>(null)
const ref = useCallback((refNode) => (node.value = refNode), [])

useSignalEffect(() => {
  if (open.value) node.value?.focus()
})
```

If this could somehow work like useEffect out of the box though that'd be awesome!

luisherranz:
@JoviDeCroock, @marvinhagemeister: any idea how to implement this? I can work on it.

JoviDeCroock:
I would say that it would work like the following, we introduce a second variable an object containing mode/timing as a key with three values

- sync (default): current behaviour 
[...]

PR #290, opened 2023-01-05
feat(signals-preact): add possibility of deferring effect execution to the next frame

JoviDeCroock:
Resolve https://github.com/preactjs/signals/issues/228

This currently adds support for `useEffect` like execution to the `useSignalEffect`, this is supported by means of a new `options` second argument, which is an object that allows you to pass in a property named `mode`. By default we will do `sync` execution which is the behavior we have seen before, but we also allow for a new type of behavior, namely `afterPaint` which is the same method used as `useEffect`.

Currently I did not really see a need to add `useLayoutEffect` but we could in the future, for React we could implement this like the following:

```js
export const useSignalEffect = (cb, opts) => {
  const [effectTrigger, setEffect] = useState(0);

  const effectFlush = useRef();
  const callback = useRef(cb);
  callback.current = cb;

  useEffect(() => {
    const notify = () => {
	if (mode === 'afterPaint') {
		setEffect((e) => e + 1)
	} else {
		eff.flush();
	}      
    }
    const eff = createFlusher(() => callback.current(), notify);
    effectFlush.current = eff.flush;
    return eff.dispose
  }, Empty);
  
  useEffect(() => {
    if (effectTrigger > 0) {
      effectFlush.current()
    }
  }, [effectTrigger]);
}
```

Do note that this would only trigger after updating the DOM-node if the relevant property is used on a DOM-node as well

[demo](https://codesandbox.io/s/cool-brown-72judb?file=/src/index.js)

### DIFF
--- a/.changeset/fair-moles-sin.md
+++ b/.changeset/fair-moles-sin.md
@@ -1,0 +1,5 @@
+---
+"@preact/signals": minor
+---
+
+Add support for delaying effect execution to the next frame
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -390,9 +390,9 @@ export function useSignalEffect(cb: () => void | (() => void), options?: EffectO
 		const execute = () => callback.current();
 		const notify = () => {
 			if (mode === 'afterPaint') {
-				afterNextFrame(execute)
+				afterNextFrame(eff.flush)
 			} else {
-				callback.current();
+				eff.flush();
 			}
 		}
 		const eff = createFlusher(execute, notify);
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -49,6 +49,17 @@ function createUpdater(update: () => void) {
 	return updater;
 }
 
+function createFlusher(compute: () => unknown, notify: () => void) {
+	let flush!: () => void;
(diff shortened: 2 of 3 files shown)


JoviDeCroock:
Closing due to inactivity

luisherranz:
@JoviDeCroock, we've added this in our codebase and it's working great so far: https://github.com/WordPress/block-interactivity-experiments/pull/226

The only problem is that we had to use the mangled properties instead:

```js
// Using the mangled properties:
// this.c: this._callback
// this.x: this._compute
// https://github.com/preactjs/signals/blob/main/mangle.json
function createFlusher(compute, notify) {
	let flush;
	const dispose = effect(function () {
		flush = this.c.bind(this);
		this.x = compute;
		this.c = notify;
		return compute();
	});
	return { flush, dispose };
}
```

luisherranz:
I changed the implementation a bit to ignore notifications that happen before the callback has been executed and thus prevent infinite loops:

- https://github.com/WordPress/gutenberg/pull/53358
- [Stackblitz demo](https://stackblitz.com/edit/vitejs-vite-zdrgfo?file=src/main.jsx,src/useSignalEffect.js)

PR #302, opened 2023-02-08
Fix signals aborting context updates

marvinhagemeister:
We noticed that a component that subscribes to a signal may block updates in a scenario where the signal didn't update, but context values did.

Authored together with @developit and @mdentremont in a paring session.

### DIFF
--- a/mangle.json
+++ b/mangle.json
@@ -61,6 +61,7 @@
       "$_subscribe": "S",
       "$_unsubscribe": "U",
       "preact": "",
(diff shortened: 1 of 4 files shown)


marvinhagemeister:
Managed to narrow the remaining issue to a reproducible test case.

marvinhagemeister:
Closing as the actual source of the problem was a bug in Preact. See https://github.com/preactjs/preact/pull/3889 .

PR #304, opened 2023-02-14
useSignal: Use 'useRef()' instead of 'useMemo()' to wrap the signal

rejhgadellaa:
This is a minor change so that `useSignal()` is a wrapper around `useRef()` instead of `useMemo()`.

More on the change:
https://mastodon.social/@rgadellaa/109863125430115425

### DIFF
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -332,7 +332,7 @@ Component.prototype.shouldComponentUpdate = function (
 };
 
 export function useSignal<T>(value: T) {
-	return useMemo(() => signal<T>(value), []);
+	return useRef(signal<T>(value)).current;
 }
 
 export function useComputed<T>(compute: () => T) {


rejhgadellaa:
Ah wait I should pbbly also update the docs as it contains a code snippet of what `useSignal()` actually does. Hold my beer...

JoviDeCroock:
I think one of the reasons why might be that [useRef uses useMemo under the hood](https://github.com/preactjs/preact/blob/master/hooks/src/index.js#L265) 😅

packages/preact/src/index.ts
```diff
@@ -332,7 +332,7 @@ Component.prototype.shouldComponentUpdate = function (
 };
 
 export function useSignal<T>(value: T) {
-	return useMemo(() => signal<T>(value), []);
```
billybimbob, line 335: I think that one important detail to note is that the `useRef` version is passing in `signal<T>(value)` as compared to `() => signal<T>(value)` for `useMemo`.

This means that whenever `useSignal` is reran from say state changes, **a new signal obj is created**, even though the new signal objs are effectively ignored.

Overall, I do think that the orginal `useMemo` is actually the better version.

rejhgadellaa: @billybimbob Hmm good point. That said, the `useMemo`-version creates a new method (`() => signal<T>(value)`) every time `useSignal` gets (re)called? (I do assume creating a signal is more work than creating a function that's never called?)

billybimbob: That's also a good point about the methods.

I guess really with either form, an object is being created in some form, either a signal object or a function.

I personally don't know how much overhead is added when creating an arrow function as compared to any other objects, but at least in my opinion, the *semantics* of creating a function that is ignored makes more sense than an actual signal value that is ignored.

rejhgadellaa: Yeah, and in this case `useRef` actually uses `useMemo` under the hood [link](https://github.com/preactjs/preact/blob/master/hooks/src/index.js#L265), so I guess this change doesn't make sense. At least not for preact (I think react's implementation of `useRef` is different, so maybe the switch from `useMemo` still makes sense there).

andrewiggins:
Per discussion, closing this since in Preact `useMemo` and `useRef` are basically the same. We may do this for React though since the implementation there is different between the two hooks.

Issue #299, opened 2023-01-28
Extend the API with set method to match the Svelte store contract


Thiagolino8:
First of all, thank you for the library!

The .subscribe() method has already been added to the library, and it was a great addition as it makes it compatible with the svelte $ syntax using the store contract, but only the getter part of the contract was added, but it would be great also take advantage of the setter's part of the contract.
With a .set() method that behaves exactly like the setter value(), svelte can identify when a store is being assigned a new value with an = sign and uses the .set() method to assign this new value
Before the .set() method:
![Code_phmWnfAz7b](https://user-images.githubusercontent.com/84819630/215228225-07c57900-4d37-45d9-aa2d-78263c9f333c.png)

After the .set() method:
![Code_cbARcr6viN](https://user-images.githubusercontent.com/84819630/215228287-85546788-79e2-4eaa-96e7-f97fb58170a1.png)

Thanks!

ryanatkn:
I made a demo of two other cases where this is compelling for Svelte users. The first is two-way binding a signal to an input primitive, and the second is binding a signal containing an object to a select element, which is a slightly magical convenience in Svelte.

https://svelte.dev/repl/f5870de21ef146708e51a524af1dfc17?version=3.58.0

Although I'm a Svelte user and I'd like this feature, I can see having a second way to do the same thing in such a tight library is a significant downside.

andrewiggins:
With [Svelte 5 Runes](https://svelte.dev/blog/runes) coming, supporting & maintaining our version of Signals in Svelte is less appealing. However, has @developit [mentioned in another comment](https://github.com/preactjs/signals/pull/325#issuecomment-1620854569), anyone can easily add support for our Signals as Svelte stores themselves with the following code:

```js
import Signal from '@preact/signals-core';

Signal.prototype.set = function(value) {
  this.value = value;
};

Signal.prototype.update = function(callback) {
  this.value = callback(this.peek());
};
```

newsve:
> With [Svelte 5 Runes](https://svelte.dev/blog/runes) coming, supporting & maintaining our version of Signals in Svelte is less appealing.

@andrewiggins There're still some reasons to still support Svelte out-of-the-box, e.g., with something like `@preact/signals-svelte`, let's call them in this thread signals-svelte: 

[...]

PR #325, opened 2023-03-19
feat(core): support `.set(val)` on a signal

JoviDeCroock:
Resolves https://github.com/preactjs/signals/issues/299

This implements `.set()` to conform to the Svelte store contract as we have already implemented `.subscribe()` this felt logical.

[Svelte Ref](https://github.com/sveltejs/svelte/blob/master/src/runtime/store/index.ts#L34)

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -286,32 +286,7 @@ Signal.prototype.subscribe = function (fn) {
 };
 
 Signal.prototype.set = function (value) {
-	const signal = this;
-	if (evalContext instanceof Computed) {
-		mutationDetected();
-	}
-	if (value !== signal._value) {
-		if (batchIteration > 100) {
-			cycleDetected();
-		}
-
-		signal._value = value;
-		signal._version++;
-		globalVersion++;
-
-		/**@__INLINE__*/ startBatch();
-		try {
-			for (
-				let node = signal._targets;
-				node !== undefined;
-				node = node._nextTarget
-			) {
-				node._target._notify();
-			}
-		} finally {
-			endBatch();
-		}
-	}
+	this.value = value
 };
 
 Signal.prototype.valueOf = function () {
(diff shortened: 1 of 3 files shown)


packages/core/src/index.ts
```diff
@@ -283,6 +285,35 @@ Signal.prototype.subscribe = function (fn) {
 	});
 };
 
+Signal.prototype.set = function (value) {
```
andrewiggins, line 288: Any reason we don't reuse the implementation of the value setter here and do `this.value = value`?

JoviDeCroock: Good point :joy:

JoviDeCroock:
Closing due to inactivity

Thiagolino8:
Any reason the branch was deleted and the functionality not incorporated?

marvinhagemeister:
Having two ways of doing the same thing felt confusing and we preferred to have one way to set signal values by default. The Signal class is exposed so folks who really want this can just patch it back in on the prototype.

Thiagolino8:
It's a shame, it would be very useful when using with Svelte following the store contract

developit:
FWIW the Svelte store contract also includes update(), so this wasn't really getting us to the point of being drop-in. It would make a lot more sense for someone to build and publish a Svelte integration that does the following:

```js
import Signal from '@preact/signals-core';

Signal.prototype.set = function(value) {
  this.value = value;
};

Signal.prototype.update = function(callback) {
  this.value = callback(this.peek());
};
```

Thiagolino8:
svelte's built-in store has an .update() method, but the store contract only asks for .subscribe() and .set(). It wouldn't really make sense to create a solution to add only the .set() method considering that the core library already has the .subscribe() method that was only inserted in the core library for compatibility with the svelte store contract, since for use outside of svelte there was already the effect() function

marvinhagemeister:
The `.subscribe()` method wasn't inserted only because of Svelte. It's useful when interfacing with imperative code too.

PR #335, opened 2023-03-30
Revert react integration to tracking current dispatcher

### LINKED ISSUE #251: Using signals package with react router results in a wierd jsx error
I tried using `@preact/signals-react` in a project with react-router 6 but I got that error which doesn't make any sense to me, the error is complaining about react-router but originates from `@preact/signals-react` and only happens when I import/use anything from `@preact/signals-react`

image

here is a link https://stackblitz.com/edit/react-router-starter-gfxkdm?file=components%2FHome.js  to reproduce the error, try commenting out the signal import/usage in Home.js and everything will work fine 

I think it could be related to the patches to JSX factories, which breaks the react-router check for the `<Route/>`  element type somehow, if I used it with the routing config approach from react-router 6.4 it works fine

### LINKED ISSUE #330: How to make Signals work with Routes
Hi, does Signal not work with `Routes` from react-router-dom. 

I notice every time I include the import Signal statement (`import { useSignal } from '@preact/signals-react';`) in a component file where I use Routes. It will throw the error below:

```
Uncaught Error: [Route] is not a <Route> component. All component children of <Routes> must be a <Route> or <React.Fragment>
```

image

```
import React, { Suspense } from 'react';
import Box1 from './Box1';
import {
  Route, Navigate, Routes,
} from 'react-router-dom';

const SignalsPattern = () => (
       <Routes>
        <Route
          path='/procedure'
          element={(
            <Suspense fallback=''>
              <Box1 />
            </Suspense>
          )}
        />
      </Routes>
);
```

andrewiggins:
Revert our react integration back to setting up signal subscription tracking by watching changes in the ReactCurrentDispatcher instead of proxying function Component calls in createElement so it works with more ecosystem libraries such as react-router-dom.

Check the comment in the source for a deeper discussion of the implementation details and choices.

Fixes #251 
Fixes #330

### DIFF
--- a/packages/react/src/index.ts
+++ b/packages/react/src/index.ts
@@ -31,6 +31,49 @@ const ProxyInstance = new WeakMap<
 	FunctionComponent<any>
 >();
 
+// Idea:
+// - For Function components: Use CurrentDispatcher to add the signal effect
+//   store to every component (kinda expensive?).
+//    - Actually we could probably skip using useSyncExternalStore and just use
+//      the effect instance directly... Ideally that'd means components that
+//      don't use any signals incur no persistent memory cost, outside of an
+//      empty call to useReducer to generate a rerender function.
+//
+//      Though maybe useSyncExternalStore makes it more concurrent mode safe? It
+//      seems that useSyncExternalStore may be efficient enough if we don't
+//      allocate more objects (aka the store). Though the
+//      `pushStoreConsistencyCheck` function would have a object per
+//      component... and then it'd have to loop through all of them to check if
+//      a store changed while rendering (if doing non-blocking work? So
+//      something concurrent related?). useSyncExternalStore probably isn't
+//      intended to be used on EVERY component.
+//
+//      Conclusion: Let's avoid useSyncExternalStore for now, and bring it if we
+//      find bugs.
+//
+//    - When it is set to a valid dispatcher (not the invalid one), start a
+//      preact/signal effect
+//    - When it is set to the invalid dispatcher (one that throws), stop the
+//      effect
+//    - Note: If a component throws, the CurrentDispatcher is reset so we should
+//      be able to clear our state
+//    - We need to store the created updater (aka Effect) for each component to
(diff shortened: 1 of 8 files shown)


packages/react/src/index.ts
```diff
@@ -163,24 +103,138 @@ function createEffectStore() {
...
+		const isExitingComponentRender =
+			currentDispatcherType === ValidDispatcherType &&
+			nextDispatcherType === ContextOnlyDispatcherType;
+
+		// Update the current dispatcher now so the hooks inside of the
+		// useSyncExternalStore shim get the right dispatcher.
+		currentDispatcher = nextDispatcher;
+		if (isEnteringComponentRender) {
+			lock = true;
+			const storeRef = nextDispatcher.useRef<EffectStore>();
```
andrewiggins, line 181: Instead of using a weakmap or tracking ReactCurrentOwner, I'm just using a good 'ol ref to store our updater/store.

Note: ReactCurrentOwner is only used for class components in the production build of React so we can't use that anyway :/

XantreGodlike:
This is really good job

packages/react/src/index.ts
```diff
@@ -163,24 +103,138 @@ function createEffectStore() {
...
+			if (storeRef.current == null) {
+				storeRef.current = createEffectStore();
 			}
+
+			const store = storeRef.current;
+			useSyncExternalStore(
+				store.subscribe,
+				store.getSnapshot,
+				store.getSnapshot
+			);
```
andrewiggins, line 191: TODO: move these hook calls into a custom hook function for hopefully better grouping in devtools

JoviDeCroock (approved):

ShivamJoker:
Hoping it fixes some of my issues as well.

andrewiggins:
Gonna go ahead and merge this now. Happy to address any feedback afterwards though. Just @-mention in any comments :)

salamaashoush:
@andrewiggins  Thanks a lot for fixing that issue :D

Issue #301, opened 2023-02-08
Bug: signals aren't reactive on lazy imported components


waelbettayeb:
Signals aren't reactive on lazy imported components on the react version. However, in the preact experiment suspense works fine.

englishisnotenglish:
same issue

agileago:
same +1

loicnestler:
Any updates on this?

ywang1724:
same +1

andrewiggins:
Could someone provide a minimal repro of the issue? Is this using react or preact package? What versions, etc. etc.

ywang1724:
> Could someone provide a minimal repro of the issue? Is this using react or preact package? What versions, etc. etc.

hi~, @andrewiggins , you can reproduce it with this [demo](https://github.com/ywang1724/signals-app-demo).

A few points:

- react@18.2.0
- @preact/signals-react@1.2.2

Steps to reproduce:

1. Run `pnpm i && pnpm dev`;
2. Visit page: http://localhost:5173, and click the button;
3. You can see this phenomenon: the signal named count, its value changes and the direct imported component follow the changes,  but the lazy imported component did not follow.

XantreGodlike:
#420 it also is not working with forward refs

mendrik:
same problem with react.signals in 2.2.0 + @preact/signals-react-transfor:  0.4.0

JoviDeCroock:
As we are pushing people towards [the Babel plugin](https://github.com/preactjs/signals/tree/main/packages/react-transform) and away from the automatic runtime I’ll close this issue.

PR #342, opened 2023-04-10
fix: fix signals are not reactive on lazy imported components.(#301)

ywang1724:


XantreGodlike:
I am not that it is correct

PR #344, opened 2023-04-12
Allow react ^16.14.0 as a peer dependency

andrewiggins:
Is there any reason we can't support ^16.14.0? [According to npmfs.com](https://npmfs.com/package/react/16.14.0/), the react 16.14.0 package is the first version that contains the `jsx-runtime` and `jsx-dev-runtime` files so it shouldn't cause any errors. Running the tests locally on 16.14.0 pass as well.

### DIFF
--- a/.changeset/fluffy-pillows-shout.md
+++ b/.changeset/fluffy-pillows-shout.md
@@ -1,0 +1,5 @@
+---
+"@preact/signals-react": patch
+---
+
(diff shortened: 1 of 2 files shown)


developit (approved):
LGTM! Took me a minute to realize you picked 16.14 due to that being the version when jsx-runtime was introduced.

andrewiggins:
Updated the description to hopefully make that clearer ☺️

PR #348, opened 2023-04-14
Refactor

AbdurRahamanAR:
Change some non-muted variables from let -> const


rschristian:
Thanks for taking the time to make a PR, but this isn't really a meaningful change. Without an improvement in behavior or readability, a PR doesn't pose any real value.

As such, I'm going to close this out.

PR #352, opened 2023-04-19
fix: Resolve full useSyncExternalStore path

rschristian:
https://twitter.com/impatienceisav1/status/1648811982059888640

It'd be nice if the React team provided any form of package exports here

### DIFF
--- a/.changeset/slow-dancers-serve.md
+++ b/.changeset/slow-dancers-serve.md
@@ -1,0 +1,5 @@
+---
+"@preact/signals-react": patch
+---
+
+Uses full file path on useSyncExternalStore import, fixing a possible resolution issue in some build tools.
--- a/packages/react/src/index.ts
+++ b/packages/react/src/index.ts
@@ -20,7 +20,7 @@ import {
 	Signal,
 	type ReadonlySignal,
 } from "@preact/signals-core";
-import { useSyncExternalStore } from "use-sync-external-store/shim/index";
+import { useSyncExternalStore } from "use-sync-external-store/shim/index.js";
 import type { Effect, JsxRuntimeModule } from "./internal";
 
 export { signal, computed, batch, effect, Signal, type ReadonlySignal };


rschristian:
May not fix the issue, despite the error message?

https://twitter.com/impatienceisav1/status/1648852878595948545

samuelgoff:
> May not fix the issue, despite the error message?
> 
> https://twitter.com/impatienceisav1/status/1648852878595948545

Hey @rschristian thanks so much for making the effort! 🙏🏼 Perhaps I'm missing a step, but simply applying this change directly inside my `node_modules` (specifically `node_modules/@preact/signals-react/src/index.ts`), then restarting my app (which `builds client and server successfully`) without making any other changes should incorporate the patch, no?

If there's something else I should be doing, please let me know. I'm happy to give it a shot.
Screenshot 2023-04-20 at 10 34 40 AM

rschristian:
@samuelgoff it looks like you're using pnpm workspaces, not the standard `node_modules` setup.

Patch `@preact/signals-react` according to the path in the error message.

samuelgoff:
> @samuelgoff it looks like you're using pnpm workspaces, not the standard `node_modules` setup.
> 
> Patch `@preact/signals-react` according to the path in the error message.

Hey @rschristian Yes, pnpm workspaces create symbolic links inside every workspace inside a monorepo, so by editing the file inside the specific workspace, I am editing the same one referenced in the error. I just traversed the path from the top-level `node_modules`, and the change I made is there.

Effectively, these files are one & the same. But on a large monorepo, it's just faster & easier to find the file referenced from among a smaller list of `node_modules`, because it's filtered to include only dependencies for that specific workspace.

Screenshot 2023-04-20 at 2 03 41 PM

rschristian:
...are you editing the file in `src/`, not `dist/` like the error message specifies?

We ship the source alongside the built for ease of debugging, but you can't edit it and see changes.

samuelgoff:
> @samuelgoff it looks like you're using pnpm workspaces, not the standard `node_modules` setup.
> 
> Patch `@preact/signals-react` according to the path in the error message.

Also, please note that this exact same error was being experienced in my [codesandbox reproduction](https://codesandbox.io/p/sandbox/vigilant-glade-bg5loj?file=%2Fpages%2Findex.tsx), which doesn't use a monorepo, but does use `pnpm` & NextJS. Hopefully this helps rule out any suspicion of the monorepo & pnpm workspaces being a factor.

samuelgoff:
> ...are you editing the file in `src/`, not `dist/` like the error message specifies?
> 
> We ship the source alongside the built for ease of debugging, but you can't edit it and see changes.
> 
> You need to edit `dist/signals.mjs`, not `src/index.ts`.

ACK! I knew it must be something stupid simple like that 🙂 Of course, now that I made the change to the `.mjs` file, it works perfectly! Thanks so much 🙏🏼

rschristian:
Yay! We've certainly all been there before, editing the wrong file and wondering why we're not seeing changes 😅

While we're waiting for a release, you could use something like [`patch-package`](https://npm.im/patch-package) to automate the patch and ensure it gets applied even across reinstalls. I hope we'll have this out soon enough though.

andrewiggins (approved):
🎉

PR #355, opened 2023-04-21
Fix React adapter in SSR and when rerendering

### LINKED ISSUE #350: Bug - v1.3.x Throws Errors when Used in React
Overview
---
Began using v1.2.2 ~3 weeks ago and it worked whether using new Signal() or signal outside of react component, or when using useSignal() within a component. Started a new project and automatically picked up v1.3.1. As soon as I began implementing signals, the app immediately broke.

Packages (using remix.run + React)
---
Remix.run - 1.1.15
Node.js - v18.12.0

The error thrown is: 

```sh
throw new Error('Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for' + ' one of the following reasons:\n' + '1. You might have mismatching versions of React and the renderer (such as React DOM)\n' + '2. You might be breaking the Rules of Hooks\n' + '3. You might have more than one copy of React in the same app\n' + 'See https://reactjs.org/link/invalid-hook-call for tips about how to debug and fix this problem.');
          ^
Error: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
See https://reactjs.org/link/invalid-hook-call for tips about how to debug and fix this problem.
```

There are also some other things related to goroutines and esbuild errors that throw:

```sh
fatal error: all goroutines are asleep - deadlock!

goroutine 1 [chan receive]:
github.com/evanw/esbuild/internal/helpers.(*ThreadSafeWaitGroup).Wait(...)
        github.com/evanw/esbuild/internal/helpers/waitgroup.go:36
main.runService.func2()
        github.com/evanw/esbuild/cmd/esbuild/service.go:127 +0x59
main.runService(0x1)
        github.com/evanw/esbuild/cmd/esbuild/service.go:181 +0x530
main.main()
        github.com/evanw/esbuild/cmd/esbuild/main.go:228 +0x21c

goroutine 34 [chan receive]:
main.runService.func1()
        github.com/evanw/esbuild/cmd/esbuild/service.go:111 +0x3e
created by main.runService
        github.com/evanw/esbuild/cmd/esbuild/service.go:109 +0x1e5

goroutine 35 [chan receive]:
main.(*serviceType).sendRequest(0xc00010dd40, {0x14a91c0, 0xc001c692c0})
        github.com/evanw/esbuild/cmd/esbuild/service.go:210 +0x110
main.runService.func3()
        github.com/evanw/esbuild/cmd/esbuild/service.go:138 +0x39
created by main.runService
        github.com/evanw/esbuild/cmd/esbuild/service.go:135 +0x353
```

Temporary Solution
---
Rollback to v1.2.2.

andrewiggins:
Updates the react adapter to handle SSR and components that call setState while rendering.

I've also added a suite of mocha test that run in NodeJS and use react-dom/server to validate our adapter works in that environment.

Note:
* The `@preact/signals-react` package can not be lazily imported. It must be imported before you call `ReactDOM.render`, `ReactDOM.hydrate`, `renderToStaticMarkup`, etc.
* Your bundler must allow overriding other modules' exports. We use this capability to wrap React's `createElement` and related methods to handle passing signals as attributes to DOM elements.

Fixes #350


packages/react/test/node/setup.js
```diff
@@ -0,0 +1,52 @@
...
+	presets: [ts, jsx, env],
+	plugins: [
+		coverage && [
+			"istanbul",
+			{
+				// TODO: Currently NodeJS tests always run against dist files. Should we
+				// change this?
+				// include: minify ? "**/dist/**/*.js" : "**/src/**/*.{js,jsx,ts,tsx}",
+			},
+		],
```
andrewiggins, line 50: Note: coverage isn't reported from this test run and I got tired and didn't set up running tests against source. If someone is feeling motivated and wants to do so, please do!

JoviDeCroock (approved):
Damn, this was some research work I reckon :o

packages/react/src/index.ts
```diff
@@ -153,15 +165,89 @@ function usePreactSignalStore(nextDispatcher: ReactDispatcher): EffectStore {
...
+Below is a state machine definition for transitions between the various
+dispatchers in React's prod build. (It does not include dev time warning
+dispatchers which are just always ignored).
+
+ENTER and EXIT suffixes indicates whether this ReactCurrentDispatcher transition
+signals we are entering or exiting a component render, or if it doesn't signal a
+change in the component rendering lifecyle (NOOP).
+
+```js
+// Paste this into https://stately.ai/viz to visualize the state machine.
```
rschristian, line 200: This is pretty darn neat

rschristian (approved):
One of those rare PRs where the comments are so good that I came away feeling like I learned a thing or two.

Those things might be semi-cursed knowledge of React internals, but hey, knowledge is knowledge! 😅

marvinhagemeister (approved):

andrewiggins:
Haha @rschristian definitely some cursed knowledge for sure 🧙🪄🔮

XantreGodlike:
There are really hard job with researching weird react internals. Really appreciate.
I am for now fan of @andrewiggins btw

PR #367, opened 2023-05-25
feat(signals-preact): useLiveSignal

### LINKED ISSUE #366: Bug?: useComputed doesn't update when given a new signal
I'm wondering if this is intended behavior or not. 

#### Steps to preroduce:
- Open https://stackblitz.com/edit/vitejs-vite-mebrre?file=src%2Fapp.jsx
- Click **Increase A**
- Observe that the large number in grams updates to reflect A
- Click **Show B in grams**
- Observe that the large number in grams _doesn't update_ to reflect B 
- Click **Increase B**
- Observe that the large number in grams _still doesn't update_  to reflect B
- Click **Increase A**
- Observe that the large number in grams _finally updates_ to reflect B

So it seems that `useComputed` can only start tracking a new signal when the previous one changes. If this is intended, it should be very well documented that you need to pass the same instance of the signal to get proper behavior.

JoviDeCroock:
Resolve #366

This adds `useLiveSignal` as a way to wrap a signal passed into a component which can change reference.

### DIFF
--- a/.changeset/tasty-timers-greet.md
+++ b/.changeset/tasty-timers-greet.md
@@ -1,0 +1,5 @@
+---
+"@preact/signals": minor
+---
+
+Add `useLiveSignal` for instances where prop-based signal references change
--- a/packages/preact/README.md
+++ b/packages/preact/README.md
@@ -67,6 +67,38 @@ function Counter() {
 }
 ```
 
+#### Signal references
+
+When you are passing signals to components and you change the reference then the `useComputed` or `useSignalEfect`
+relying on these won't notice the difference, for that we have `useLiveSignal`.
+
+```js
+import { useComputed, useLiveSignal, useSignal } from '@preact/signals';
+
+// The signal representing props.date can change
+// when i.e. someone presses a butotn
+function Todo({ date }) {
+  const liveDateSignal = useLiveSignal(date);
+  const formattedDate = useComputed(() => formatDate(liveDateSignal.value.value));
+  return <span>{formattedDate}</span>;
+}
+
+function App() {
+  const dateA = useSignal(0);
+  const dateB = useSignal(1);
+  const [on, setOn] = useState(true);
+
+  return (
+    <main>
+      <button onClick={() => setDisplayed(!on)}>
+        Toggle
+      </button>
+    	<Todo date={on ? dateA : dateB} />
+    </main>
+  );
+}
+```
+
 ### Rendering optimizations
 
 The Preact adapter ships with several optimizations it can apply out of the box to skip virtual-dom rendering entirely. If you pass a signal directly into JSX, it will bind directly to the DOM `Text` node that is created and update that whenever the signal changes.
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -352,6 +352,17 @@ export function useSignalEffect(cb: () => void | (() => void)) {
 }
 
 /**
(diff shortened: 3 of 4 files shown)


andrewiggins:
Oohh interesting. I hadn't thought of this kind of use case before. Thanks for bringing this up!

One thought about this new hook is that it requires the child component (`Todo` in the Readme example) to handle if its parent *might* pass in a new signal object as a prop. It seems to me all components might want to call `useLiveSignal` instead of `useSignal` to handle getting a new signal object since it can't know how it might be called. Could the number of `.value`s you have to deference depend on how many calls to `useLiveSignal` your parents have invoked on that signal?

I wonder if there are other patterns we could instead encourage to handle the scenario mentioned in the Readme. It appears there are two signals that hold  a value for which there is another "current" value. Could this "current" value be a signal, i.e. `currentDate`? The current date depends on the `on` React state so wherever it is updated the `currentDate` signal would also need to be updated. So the button would become something like `<button onClick={() => { setDisplayed(!on); currentDate.value = !on ? dateA : dateB;}>`.

Though having to change state twice (once for React state and another for signal state) isn't great. It'd be nice to just do `useComputed(() => on ? dateA : dateB);` or something but since `on` is React state and not a signal, this computed value won't rerun when `on` changes. This problem makes me think another issue we have is mixing & synchronizing React state and Signal state. I wonder if we should add a hook to help that situation to encourage people to not change signal instances and instead create computed signals to represent the final value. Perhaps something like `const [ state, setState, stateSignal] = useSignalState()` which looks and behaves like `React.useState` but also exposes the state as a signal that can be used in things like `useComputed` (though synchronizing the UI when some UI updates synchronously (UI powered by `stateSignal`) and others update asynchronously (UI powered by `setState`) (maybe using `useSyncExternalStore` could help us here?

I'd be curious if there are other situations where switching signal instances in rendering may be relevant cuz perhaps I need to dwell on it a bit more to better understand it use cases.

Anyways, that's a lot of thoughts. Would love to hear what you think!

XantreGodlike:
Why we cannot do it like that?
```js
import { useComputed, useLiveSignal, useSignal } from '@preact/signals';

// The signal representing props.date can change
// when i.e. someone presses a butotn
function Todo({ date }) {
  const formattedDate = useComputed(() => formatDate(liveDateSignal.value));
  return <span>{formattedDate}</span>;
}

function App() {
  const dateA = useSignal(0);
  const dateB = useSignal(1);
  const on = useSignal(true)

  return (
    <main>
      <button onClick={() => {on.value = !on.peek()}}>
        Toggle
      </button>
    	<Todo date={useComputed(() => on.value ? dateA.value : dateB.value)} />
    </main>
  );
}
```

XantreGodlike:
I think in most cases when we are writing component powered by signals we should stick with then and do not use it without meaningful reason

JoviDeCroock:
@andrewiggins I see your point, I honestly dislike the `.value.value` consequence of having to access it this way. The main problem here becomes that the reference to the signal isn't safe so maybe just incorporating this in `useSignal` is the most intuitive way of incorporating this however that probably would need to be stateful i.e. rather than `useSignal` using `useMemo` to just create a signal once it would need to create it in state and check whether it's referentially equal on each invocation. WDYT?

XantreGodlike:
@JoviDeCroock maybe it can be solved with eslint plugin for signals that will enforce stable reference to signals

PR #371, opened 2023-06-09
Add tests verifying the behavior of nested effects and computeds

andrewiggins:
Specifically tests creating and passing computed between components

### DIFF
--- a/packages/preact/test/index.test.tsx
+++ b/packages/preact/test/index.test.tsx
@@ -189,6 +189,70 @@ describe("@preact/signals", () => {
 			rerender();
 			expect(spy).to.be.calledOnce;
 		});
+
+		it("should not subscribe to computed signals only created and not used", () => {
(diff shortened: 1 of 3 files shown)


packages/preact/test/index.test.tsx
```diff
@@ -189,6 +189,70 @@ describe("@preact/signals", () => {
 			rerender();
 			expect(spy).to.be.calledOnce;
 		});
+
+		it("should not subscribe to computed signals only created and not used", () => {
```
andrewiggins, line 193: @developit Does this test capture the behavior we've spoken about in some of our meetings re: creating and passing down computed values. Just figured I'd add a test to ensure the core integration libraries have the behavior we expect.

marvinhagemeister (approved):

JoviDeCroock (approved):

PR #372, opened 2023-06-15
Reorganize signals-react package

andrewiggins:
Reorganize the signals-react package in preparation for some future explorations to improve the integration with React.


JoviDeCroock (approved):

packages/react/runtime/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "main": "dist/runtime.js",
+  "module": "dist/runtime.module.js",
+  "unpkg": "dist/runtime.min.js",
+  "types": "dist/index.d.ts",
+  "source": "src/index.ts",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "browser": "./dist/runtime.module.js",
+      "umd": "./dist/runtime.umd.js",
```
rschristian, line 15:
```suggestion
      "umd": "./dist/runtime.min.js",
```

I recognize you copied from our other packages, but they look to have been misconfigured too.

Alternatively, we could just remove it. We have yet to see a complaint from this broken configuration in the other packages and I don't know of any tools that will actually consume `exports.umd`.

packages/react/runtime/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "name": "@preact/signals-react-runtime",
+  "description": "Sub package for @preact/signals-react that contains the pure runtime functions",
+  "private": true,
+  "amdName": "reactSignalsRuntime",
+  "main": "dist/runtime.js",
+  "module": "dist/runtime.module.js",
+  "unpkg": "dist/runtime.min.js",
+  "types": "dist/index.d.ts",
+  "source": "src/index.ts",
+  "exports": {
```
rschristian, line 11: I'm guessing you don't plan on this being a private, build-time only package forever, given the full slew of package formats?

rschristian (approved):

packages/react/runtime/package.json
andrewiggins: Oh good catch. I'll fix them.

rschristian: Alternatively, #373. Not sure which is better

packages/react/runtime/package.json
andrewiggins: The sub packages of preact (e.g. [preact/hooks](https://github.com/preactjs/preact/blob/master/hooks/package.json)) have all formats also defined so I did the same.

I haven't fully decided on if this'll be its own package but its not out of the question.

packages/react/runtime/package.json
andrewiggins: Oh whoops! Just saw that. We can do that.

packages/react/runtime/package.json
rschristian: Ah, fair enough!

For some reason I was thinking that this would be inlined into the React package. My bad!

PR #373, opened 2023-06-20
refactor: Remove non-functional `package.json#exports.umd`

rschristian:
In all three of the packages, `package.json#exports.umd` has been referencing the incorrect path for the UMD bundles since release (should've been `x.min.js`).

However, seeing as how I can't see a single complaint relating to this, and I'm not aware of any tools that actually consume it (would love to know if there are tools that do), I elected to remove the entries. Saves us from making a breaking change in the future if we were to alter that entry in any way.

### DIFF
--- a/.changeset/chilled-cheetahs-grin.md
+++ b/.changeset/chilled-cheetahs-grin.md
@@ -1,0 +1,7 @@
+---
+"@preact/signals-core": patch
+"@preact/signals": patch
+"@preact/signals-react": patch
+---
+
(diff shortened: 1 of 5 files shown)


JoviDeCroock (approved):

andrewiggins (commented):
Can you also update the react/runtime/package.json as well? I missed this PR when I completed that one lol

rschristian:
Can do! I should've mentioned this on that PR, but evidently forgot to.

andrewiggins (approved):

PR #375, opened 2023-06-23
Add package `@preact/signals-react-transform`

andrewiggins:
This PR adds a new package `@preact/signals-react-transform` that provides a Babel plugin to transform functions to observer/track what signals are used when the function runs. This plugin is the start to an alternate React integration for signals that does not hook into React internals. Once our approach here is finalized and published, I'll update the `@preact/signals-react` package to reference this and provide various options for using signals in React.

Check [the test file](https://github.com/preactjs/signals/pull/375/files#diff-95ee027f0e90ee403bb02268d8b0de381d341a38fd0632ac1ec3838a43bbf365R66-R77) for examples of what the transformation looks like.

I'm currently exploring two different kinds of transformations in this implementation to see which one works better. One just adds a call to `useSignals()` to the top of the component and relies on the next invocation of `useSignals` or a microtick to stop tracking signals. The other wraps the component body in a `try/finally` to start and stop tracking signals. I'm currently leaning towards making the `try/finally` the default cuz it leaves less room for tricky edge cases that may lead to non-obvious behavior and rerenders.

However, the implementation of `useSignals` is the same either way. We'll always setup the microtick to clean up any outstanding effects to prevent any loose ends regardless of which transformation is used. So users can manually  drop `useSignals` into a component and get basic functionality.


karma.conf.js
```diff
@@ -283,7 +287,7 @@ module.exports = function (config) {
 			jsx: "preserve",
 
 			// esbuild options
-			target: downlevel ? "es5" : "es2015",
+			target: downlevel ? "es5" : "es2019",
```
andrewiggins, line 290: Whoops - will undo this. Did this to not downlevel async/await while debugging tests

packages/react-transform/README.md
```diff
@@ -0,0 +1,119 @@
+# Signals React Transform
```
andrewiggins, line 1: Unsure what's the right level of detail here. Would love other's thoughts.

packages/react-transform/src/index.ts
```diff
@@ -0,0 +1,350 @@
...
+	PluginObj,
+	PluginPass,
+	NodePath,
+	template,
+} from "@babel/core";
+import { isModule, addNamed, addNamespace } from "@babel/helper-module-imports";
+
+// TODO:
+// - how to trigger rerenders on attributes change if transform never sees
+//   `.value`?
```
andrewiggins, line 13: I still need to build the signal attribute handling in our jsx handler

packages/react-transform/src/index.ts
```diff
@@ -0,0 +1,350 @@
...
+					);
+				},
+			},
+
+			ArrowFunctionExpression: {
+				// TODO: Consider alternate implementation, where on enter of a function
+				// expression, we run our own manual scan the AST to determine if the
+				// function uses signals and is a component. This manual scan once upon
+				// seeing a function would probably be faster than running an entire
+				// babel pass with plugins on components twice.
```
andrewiggins, line 214: For anyone with more experience with babel transforms - if I want to know if a function body contains JSX to know whether to perform my transformation, is it better to do a local visitor pass inside this ArrowFunctionExpression handler to observe that (only running my local visitor), and then transform? Or to do the observation in the larger babel visitor pass and then transform after, which triggers and additional visitor pass with all plugins?

marvinhagemeister: That's a good question. I've never thought about this.

packages/react-transform/README.md
marvinhagemeister: I think it's perfectly fine. We could theoretically shorten the intro paragraph about what signals are by just pointing to the blog post. But then again that I think leaving as is, is fine too.

andrewiggins:
TODO for myself: Add some E2E babel transform tests, i.e. transform code and run it.

XantreGodlike:
What if someone calls `useSignals` twice? Or babel will add additional one?

XantreGodlike:
What about cases when signals used inside hooks that component uses? How to handle this case?

andrewiggins:
> What if someone calls useSignals twice? Or babel will add additional one?

andrewiggins:
> What if someone calls useSignals twice? Or babel will add additional one?
`useSignals` can be called multiple times in a component. It'll recreate the effect and other relevant tracking machinery but that's okay. The previous effect will be cleaned up and any signals used after the additional calls to useSignals will still trigger rerenders properly.

> What about cases when signals used inside hooks that component uses? How to handle this case?
Hooks that use signals are transformed to call `useSignals` at the top of the function. In this situation, we are relying on the fact that calling `useSignals` multiple times is safe to do.

packages/react-transform/src/index.ts
andrewiggins: Will do this in a follow up PR. this one is big enough 😅

packages/react-transform/src/index.ts
```diff
@@ -0,0 +1,369 @@
...
+
+function hasLeadingOptInComment(path: NodePath) {
+	return hasLeadingComment(path, optInCommentIdentifier);
+}
+
+function hasLeadingOptOutComment(path: NodePath) {
+	return hasLeadingComment(path, optOutCommentIdentifier);
+}
+
+function isOptedIntoSignalTracking(path: NodePath | null): boolean {
```
andrewiggins, line 91: Here and in `isOptedOutOfSignalTracking`, I'm climbing up the AST to find the most logical place for a comment to live. For example supporting a use case such as:
```js
/** @trackSignals */
export default function useCustomHook() {}
```

andrewiggins (commented):
Okay! I believe this is now ready for review :tada:

packages/react-transform/src/index.ts
```diff
@@ -0,0 +1,369 @@
...
+		default:
+			return false;
+	}
+}
+
+function isComponentFunction(path: NodePath<FunctionLike>): boolean {
+	return (
+		fnNameStartsWithCapital(path) && // Function name indicates it's a component
+		getData(path.scope, containsJSX) === true && // Function contains JSX
+		path.scope.parent === path.scope.getProgramParent() // Function is top-level
```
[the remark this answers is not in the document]
andrewiggins: Yea, it's true. I'm thinking HoC will need to manually opt-in using `/** @trackSignals */` or by calling `useSignals()` directly in their wrapper function.

It is harder to detect HoCs and know where to put the useSignals call so for now I'm explicitly skipping them. I should add a note to the Readme calling this out for people who want to use signals in higher order components.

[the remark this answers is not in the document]
XantreGodlike: I think it a good idea to mention edge cases in readme

packages/react-transform/test/node/index.test.tsx
```diff
@@ -0,0 +1,835 @@
...
+
+			const expectedOutput = `
+				import { useSignals as _useSignals } from "@preact/signals-react/runtime";
+				const MyComponent = () => {
+					var _effect = _useSignals();
+					try {
+						signal.value;
+						return <div>Hello World</div>;
+					} finally {
+						_effect.f();
```
andrewiggins, line 75: Any opinions on the name of this method `f`? It's short for `finishEffect` to stop the current effect.

I've changed the return of `useSignals` from an method to an object that can hang other stuff of like the dispose symbol for using with `using` which this PR also includes support for.

packages/react-transform/src/index.ts
JoviDeCroock: You can use something like [this](https://github.com/preactjs/prefresh/blob/main/packages/babel/src/index.mjs#L168-L183) on a VariableDeclaration to see whether they are used as JSX

JoviDeCroock (approved):
This looks amazing

packages/react-transform/src/index.ts
andrewiggins: Oh interesting. IIUC, this code looks to see if a particular identifier is used as JSX somewhere in the same file?

JoviDeCroock: ScopeBindings should work across files afaik

marvinhagemeister (approved):

Issue #378, opened 2023-06-30
Proposal: `untracked` api function


betula:
### Untracked

Necessary to implement a function that provides a possibility for disabling the tracking of dependencies.

It has already been implemented in Angular Signals and MobX.

> Execute an arbitrary function in a non-reactive (non-tracking) context. The executed function can, optionally, return a value.
>  Angular16 docs [`untracked`](https://angular.io/api/core/untracked)

> Runs a piece of code without establishing observers.
>  MobX docs [`untracked`](https://mobx.js.org/api.html#untracked)

**What's the reason to use it?**

For reading **one signal** without dependency creation the `peek` method already implement.
[Reading signals without subscribing to them](https://preactjs.com/guide/v10/signals/#reading-signals-without-subscribing-to-them)

> On the rare occasion that you need to write to a signal inside [`effect(fn)`](https://preactjs.com/guide/v10/signals/#effectfn), but don't want the effect to re-run when that signal changes, you can use `.peek()` to get the signal's current value without subscribing.

But it is not enough.

`untracked` function provides the possibility for the `reaction` pattern:

```javascript
function reaction(sender, listener) {
  effect(() => {
    const value = sender();
    untracked(() => {
      listener(value);
    });
  }
}
```

In that pattern, you have a way to control your dependencies. The reaction will collect dependencies only from the sender function, and react only on them.

Many thanks to you, 
your signals implementation is amazing, 
I think the best at the moment.

XantreGodlike:
I really need this functionality, too. 
In case of compositibility and easibility of usage, in some cases it's easier to use solidjs like `Accessor`
`type Accessor<T> = () => T`
Because its easy way to define lazy computation that will executed on demand. And you don't need to rely on rules of hooks. But for more flexible usage of this pattern, i need a way to turn off deps tracking on consumer side
![image](https://github.com/preactjs/signals/assets/57757211/c8fc9024-2e64-45af-a732-e0a3a5b85c78)

XantreGodlike:
#380

PR #380, opened 2023-07-05
Added `untracked` function

XantreGodlike:
Resolves https://github.com/preactjs/signals/issues/378

As mentioned in the proposal this allows us to write effects that essentially aren't re-executed by the signals that get accessed in the function body of `untracked`.

Allowing functionality like

```js
function reaction(sender, listener) {
  effect(() => {
    const value = sender();
    untracked(() => {
      listener(value);
    });
  }
}
```

### DIFF
--- a/.changeset/dirty-geese-learn.md
+++ b/.changeset/dirty-geese-learn.md
@@ -1,0 +1,5 @@
+---
+"@preact/signals-core": minor
+---
+
+Added `untracked` api, to more flexible control over dependency tracking. More info in README.md
--- a/.changeset/dirty-geese-learn.md
+++ b/.changeset/dirty-geese-learn.md
@@ -2,4 +2,4 @@
 "@preact/signals-core": minor
 ---
 
-Added `untracked` api, to more flexible control over dependency tracking. More info in README.md
+Add `untracked` function, this allows more granular control within `effect`/`computed` around what should affect re-runs.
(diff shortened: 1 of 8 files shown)


XantreGodlike:
@andrewiggins can you review or say why its shouldn't be added?)

andrewiggins:
Hey! One question that comes to mind for me is what use case does this help that use `const value = signal.peek()` doesn't work for?

XantreGodlike:
Some reasons is specified in proposal linked to the PR.
But shortly: its useful for creating utils that works with callbacks. Most of reactive runtime provides `untracked`, so it's common pattern, that allows to manipulate reactivity not in user side, but in consumer side. 
Yes, I can emulate untracked with computed, but it has too much overhead IMHO. 
```js
const untracked = (callback) => computed(callback).peek()
```

andrewiggins:
Oh whoops! I missed the linked proposal. Lemme take a look and mull it over with the team

XantreGodlike:
I think this thing are essential for writing Store-like primitives and dealing with it without tricks

XantreGodlike:
@andrewiggins when you will return with feedback?

andrewiggins:
I have some vacation coming up so will be away for a week or two. Though if other core team members have thoughts, no need to block on me

XantreGodlike:
@JoviDeCroock what do you think about `untracked`?

JoviDeCroock:
This looks good to me, tests seem to be failing and needs a changeset

XantreGodlike:
> This looks good to me, tests seem to be failing and needs a changeset

Fixed

JoviDeCroock (approved):

betula:
Thanks a lot👍

Issue #323, opened 2023-03-14
This JSX tag's 'children' prop expects a single child of type 'ReactNode', but multiple children were provided


childrentime:
```tsx
import { signal } from "@preact/signals-react";

const count = signal(0);
function App(){

return  <button
          onClick={() => {
            count.value = count.value + 1;
          }}
        >
          count is {count}
        </button>
}
```

just a simple demo code will throw typescript error, maybe we should export a new jsx type to satisfy jsxImportSource option?

rschristian:
This is due to React 18's newer, more strict children types. To fix this, you can simply wrap the content in a fragment:

```jsx
import { signal } from "@preact/signals-react";

const count = signal(0);
function App() {
    return (
        <button onClick={() => count.value = count.value + 1}>
          <>count is {count}</>
        </button>
    )
}
```

Patching React's types is not a good idea, no, nor is it specific to this library. Any library which uses objects within JSX will likely have this issue.

christowiz:
@rschristian It would be nice if this was add to the documentation and code examples so other people wouldn't have to search Github issues to figure out the problem.

rschristian:
If you have a suggestion for how to improve the docs, PRs are always very welcome.

christowiz:
@rschristian Docs PR: https://github.com/preactjs/signals/pull/385

PR #385, opened 2023-07-10
docs: add React fragment note

christowiz:
There is missing documentation regarding the TypeScript error in React when the signal object is used directly in JSX.  This requires wrapping the object in a React Fragment (see https://github.com/preactjs/signals/issues/323).

This PR adds a note about the React Fragment requirement.  The Preact 'Render Optimization' content was copied with changes to code example and added note.  This updated code sample was tested and working the correctly with React.

https://codesandbox.io/s/preact-signals-in-react-test-mzfv8x\?file\=/src/LogCount.tsx

### DIFF
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -66,6 +66,36 @@ function Counter() {
 }
 ```
 
+### Rendering optimizations
+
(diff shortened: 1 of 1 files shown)


rschristian (approved):
Thanks!

To any other core members: is there a reason we skipped listed the `Text` node optimization originally?

JoviDeCroock (approved):

Issue #383, opened 2023-07-06
Difference in behaviour of `@preact/signals` and `@preact/signals-react`


XantreGodlike:
Computed that return jsx invalidates
Expected behariour: layout updates when computed updates.
Current behaviour: layout stays the same

If computed returns jsx - reactivity stops working(preact). I don't know is current behaviour on react is bug or feature, but for me its more like feature. I don't know any reason why computed cannot return jsx. 

![image](https://github.com/preactjs/signals/assets/57757211/1a6a5816-7a0b-4d60-816f-36098ea2650f)

Repro:
https://stackblitz.com/edit/vitejs-vite-o1wsux?file=src%2Fapp.tsx

mikerob215:
@XantreGodlike this example will work if you add the `.value` prop to the computeds when you pass them to jsx. The issue isn't that computeds can't return jsx, it's that when you pass the signal directly into jsx, the preact adapter will try to set a text node directly, I think you'd need the component to fully rerender to display jsx. I don't know or use the react adapter so I can't explain the behavior there but I think it might be because the react implementation just does a rerender rather than try to optimize, something like that.

XantreGodlike:
I founded source of issue. In preact adapter we are setting up updater callback for `string`, but are not checking this it's really will be string

https://github.com/preactjs/signals/blob/9d0cb672b7aa69b8afa22d9b57a7f51de838e2cf/packages/preact/src/index.ts#L84

I think it can be solved this way
```typescript
	const s = useMemo(() => {
		// mark the parent component as having computeds so it gets optimized
		let v = this.__v;
		while ((v = v.__!)) {
			if (v.__c) {
				v.__c._updateFlags |= HAS_COMPUTEDS;
				break;
			}
		}

		const defaultUpdaterCallback = this._updater!._callback;
		const stringUpdaterCallback = () => {
			(this.base as Text).data = s.peek();
		}; 

		return computed(() => {
			let data = currentSignal.value;
			let s = data.value;
			const result = s === 0 ? 0 : s === true ? "" : s || "";
			 
			if (typeof result === 'object') {
				this._updater!._callback = defaultUpdaterCallback;
			} else {
				// Replace this component's vdom updater with a direct text one:
				this._updater!._callback = stringUpdaterCallback;
			}
			return result;
		});
	}, []);

```

XantreGodlike:
I founded source of issue. In preact adapter we are setting up updater callback for `string`, but are not checking this it's really will be string

[...]

PR #387, opened 2023-07-15
Removed difference in behaviour of @preact/signals and @preact/signal…

XantreGodlike:
Resolves #https://github.com/preactjs/signals/issues/383

When we return a VNode in a signal/computed we would incorrectly try to treat this as a text-node, which would result in us incorrectly triggering changes

### DIFF
--- a/.changeset/tender-pianos-love.md
+++ b/.changeset/tender-pianos-love.md
@@ -1,0 +1,5 @@
+---
+"@preact/signals": minor
+---
+
+Removed difference in behaviour between adapters. Now signals inserted into jsx can display not only text but other jsx as well
--- a/.changeset/tender-pianos-love.md
+++ b/.changeset/tender-pianos-love.md
@@ -2,4 +2,4 @@
 "@preact/signals": minor
 ---
 
-Removed difference in behaviour between adapters. Now signals inserted into jsx can display not only text but other jsx as well
+Removed difference in behaviour between adapters, signals that use a JSX value will correctly re-render the whole component rather than attempting the JSX-Text optimization.
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -80,10 +80,22 @@ function Text(this: AugmentedComponent, { data }: { data: Signal }) {
 			}
 		}
 
+		const defaultUpdaterCallback = this._updater!._callback;
+
 		// Replace this component's vdom updater with a direct text one:
-		this._updater!._callback = () => {
+		const stringOrJSXUpdaterCallback = () => {
+			const node = this.base;
+			// if we need to update the text directly, we should ensure that current node is a text node
+			// if it is not, we should promise preact to create it via rerendering the whole component
+			if (typeof s.peek() === "object" || node?.nodeType !== 3) {
+				defaultUpdaterCallback();
+
+				return;
+			}
+
 			(this.base as Text).data = s.peek();
 		};
+		this._updater!._callback = stringOrJSXUpdaterCallback;
 
 		return computed(() => {
 			let data = currentSignal.value;
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -7,6 +7,7 @@ import {
 	effect,
 	Signal,
 	type ReadonlySignal,
+	untracked,
 } from "@preact/signals-core";
 import {
 	VNode,
@@ -18,7 +19,15 @@ import {
 	AugmentedElement as Element,
 } from "./internal";
 
-export { signal, computed, batch, effect, Signal, type ReadonlySignal };
+export {
+	signal,
+	computed,
+	batch,
+	effect,
+	Signal,
+	type ReadonlySignal,
+	untracked,
+};
 
 const HAS_PENDING_UPDATE = 1 << 0;
 const HAS_HOOK_STATE = 1 << 1;
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -62,7 +62,7 @@ function createUpdater(update: () => void) {
  * A wrapper component that renders a Signal directly as a Text node.
  * @todo: in Preact 11, just decorate Signal with `type:null`
  */
-function Text(this: AugmentedComponent, { data }: { data: Signal }) {
+function SignalValue(this: AugmentedComponent, { data }: { data: Signal }) {
 	// hasComputeds.add(this);
 
 	// Store the props.data signal in another signal so that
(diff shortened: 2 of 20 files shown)


XantreGodlike:
@andrewiggins i think it has no performance overhead in case of rendering string into Text nodes directly, but adds oprtunity to render jsx inside signal. 
Researching `preact` source code was fun.

There are oprtunity when chaning signal type `jsx -> string`, to directly create Text node and replace it in `this.base` and `this.__v.__e`, but i am not sure will it give real performance boost

XantreGodlike:
I will add `changeset` after review

XantreGodlike:
@JoviDeCroock please check this PR 🙏🙏🙏

packages/preact/test/index.test.tsx
```diff
@@ -131,6 +131,93 @@ describe("@preact/signals", () => {
...
+			{
+				const text = scratch.firstChild!.firstChild!;
+
+				expect(text.textContent).to.equal("test");
+				expect(text).to.be.an.instanceOf(HTMLSpanElement);
+				expect(text).to.have.property("firstChild").that.is.an.instanceOf(Text);
+			}
+			sig.value = "a";
+			rerender();
+			{
```
JoviDeCroock, line 193: Why the random curlies? 😅

packages/preact/test/index.test.tsx
```diff
@@ -418,38 +505,44 @@ describe("@preact/signals", () => {
 		});
 	});
 
-	describe('hooks mixed with signals', () => {
-		it('signals should not stop context from propagating', () => {
-			const ctx = createContext({ test: 'should-not-exist' });
+	describe("hooks mixed with signals", () => {
```
JoviDeCroock, line 508: Styling change is probably unintended?

packages/preact/src/index.ts
```diff
@@ -80,9 +80,20 @@ function Text(this: AugmentedComponent, { data }: { data: Signal }) {
 			}
 		}
 
+		const defaultUpdaterCallback = this._updater!._callback;
+
 		// Replace this component's vdom updater with a direct text one:
 		this._updater!._callback = () => {
```
JoviDeCroock, line 86: We could also stop changing the `_callback` when we see that the signal isn't a primitive value. In that aspect the name of this function might be misleading as we might need to just call it `SignalValue` or something and opt in to these optimizations selectively

packages/preact/test/index.test.tsx
XantreGodlike: Yes, but it's was formatted automatically, I think someone was not run prettier

packages/preact/test/index.test.tsx
XantreGodlike: Because I declaring text in other block too

packages/preact/src/index.ts
XantreGodlike: Ok will rename it. So I think it should be renamed in react adapter too

XantreGodlike (commented):
Thanks for review

packages/preact/src/index.ts
JoviDeCroock: Renaming isn't the only course of action here I feel, we can also just conditionally override the callback rather than adding a check in the altered callback

packages/preact/test/index.test.tsx
JoviDeCroock: `let text = ` can do that as well

XantreGodlike: I know, but i dislike mutations without performance benefits). Will, replace with `let`)

XantreGodlike:
Fixed comments @JoviDeCroock

packages/preact/src/index.ts
```diff
@@ -144,10 +160,11 @@ hook(OptionsTypes.RENDER, (old, vnode) => {
 
 		updater = component._updater;
 		if (updater === undefined) {
-			component._updater = updater = createUpdater(() => {
-				component._updateFlags |= HAS_PENDING_UPDATE;
-				component.setState({});
-			});
+			// @ts-expect-error checking symbol in object
+			const isText = !!vnode.type?.[__isTextSymbol];
```
JoviDeCroock, line 164: let's keep it as before but just perform a condition around the useMemo

XantreGodlike: fixed

packages/preact/src/index.ts
```diff
@@ -80,8 +80,13 @@ function Text(this: AugmentedComponent, { data }: { data: Signal }) {
...
-		// Replace this component's vdom updater with a direct text one:
 		this._updater!._callback = () => {
+			if (isValidElement(s.peek()) || this.base?.nodeType !== 3) {
+				this._updateFlags |= HAS_PENDING_UPDATE;
+				this.setState({});
+				return;
+			}
+
 			(this.base as Text).data = s.peek();
 		};
```
JoviDeCroock, line 91: Wouldn't this be equivalent and maybe slightly easier to read, this is what I meant in my comment

```suggestion
        if (this.base?.nodeType === 3 || !isValidElement(s.peek())) {
        	this._updater!._callback = () => {
				(this.base as Text).data = s.peek();
		 	};
        }

```

JoviDeCroock:
This also needs a changeset and probably a rebase/merge so the netlify checks start succeeding

packages/preact/src/index.ts
XantreGodlike: No, this code throws, because: We cannot read `s` value before initializing.
And it will not work the same way, because:
1. While first rendering this.base will never exist, 
2. this.base.nodeType and s.peek() can change while lifetime of component as well

XantreGodlike:
@JoviDeCroock please check)

.changeset/tender-pianos-love.md
```diff
@@ -0,0 +1,5 @@
+---
+"@preact/signals": minor
+---
+
+Removed difference in behaviour between adapters. Now signals inserted into jsx can display not only text but other jsx as well
```
JoviDeCroock, line 5:
```suggestion
Removed difference in behaviour between adapters, signals that use a JSX value will correctly re-render the whole component rather than attempting the JSX-Text optimization.
```

packages/preact/src/index.ts
```diff
@@ -80,8 +80,13 @@ function Text(this: AugmentedComponent, { data }: { data: Signal }) {
 			}
 		}
 
-		// Replace this component's vdom updater with a direct text one:
 		this._updater!._callback = () => {
+			if (isValidElement(s.peek()) || this.base?.nodeType !== 3) {
```
JoviDeCroock, line 84: `isValidElement` would trigger for states that are created by means of `createElement('hi')` which is a valid textNode that we could perform the optimised render on so it will be safer to just check for `this.base.nodeType`

```suggestion
			if (this.base?.nodeType !== 3) {
```

packages/preact/src/index.ts
```diff
@@ -1,4 +1,4 @@
-import { options, Component } from "preact";
+import { options, Component, isValidElement } from "preact";
```
JoviDeCroock, line 1: RE other comment

```suggestion
import { options, Component } from "preact";
```

JoviDeCroock (approved):

packages/preact/src/index.ts
XantreGodlike: createElement('hi') will produce this -> `<hi></hi>`, is not it?

JoviDeCroock: Ah true internally we started using createVNode for this, in Preact a nullish type can be an individual text node hence the commwnt

PR #405, opened 2023-09-01
feat(core): add unique identifier for signals

### LINKED ISSUE #402: Detect a signal without tightly binding to a version of the library
We're working on a way to write a Lit html template that will automatically handle signals. The core of the implementation is:

```js
  values.map((v) => v instanceof Signal ? watch(v) : v)
```

However, this will not work properly if a user has multiple versions of `@preact/signals-core` in their node_modules tree, as it will only detect whatever version of Signal was imported in our library. Ideally, people wouldn't end up with multiple versions of signals in their node_modules of course, but in practice we find that users end up with multiple versions in their node_modules even within a single major version.

What we'd like is something like this to work:

```js

const brand = Symbol.for('preact-signals');

...

   values.map((v) => v?.[brand] ? watch(v) : v)
```

JoviDeCroock:
fixes #402

This allows for multiple versions of signals to be in a bundle and still be properly identified as signals.

### DIFF
--- a/.changeset/wild-spies-glow.md
+++ b/.changeset/wild-spies-glow.md
@@ -1,0 +1,5 @@
+---
+"@preact/signals-core": minor
+---
+
+Add unique identifier to every `Signal`, this will be present on the `type` property of a Signal coming from either `signal()` or `computed()`
--- a/mangle.json
+++ b/mangle.json
@@ -5,7 +5,11 @@
   },
   "minify": {
     "mangle": {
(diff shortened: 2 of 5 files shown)


marvinhagemeister (approved):
Great idea. Would make checking for signals a bit easier 👍

JoviDeCroock:
@marvinhagemeister do you agree with `brand` as the property name, had to change it due to `type` conflicting with ReactElement

marvinhagemeister:
@JoviDeCroock brand is fine. Fits in line with the whole "branded types" idea

PR #391, opened 2023-07-20
Upgrade to pnpm v8

andrewiggins:
It appears some of our Netlify builds (e.g. in PR #375) are failing due to pnpm version mismatch. This PR upgrades our version to the latest (v8) to fix them.

PR #431, opened 2023-11-05
Fix CI (obsolete lockfile)

XantreDev:
Updated lockfile to fix CI

Partially based on #391

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,172 +1,237 @@
-lockfileVersion: 5.4
+lockfileVersion: '6.0'
+
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
 
 patchedDependencies:
-  microbundle@0.15.1:
-    hash: yvstdq4ikeml4yz3a6bi3bgrvu
-    path: patches/microbundle@0.15.1.patch
   '@babel/plugin-transform-typescript@7.19.1':
     hash: tiqrfntt5y3ned567j2lekmz2i
     path: patches/@babel__plugin-transform-typescript@7.19.1.patch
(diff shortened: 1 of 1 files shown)


XantreGodlike:
@robertknight I've added ci for target branch

.github/workflows/compressed-size.yml
```diff
@@ -8,18 +8,17 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
```
rschristian, line 11: Following the theme, please don't alter things unless absolutely necessary.

rschristian (dismissed):
This should be targetting `main`, not the one-off branch (that I'm still not sure we'll support a release for). Everything in the workflow files looks like it should be reverted.

.github/workflows/compressed-size.yml
XantreGodlike: I think CI is needed to ensure that all works fine in the branch, also it's useful to release new version with changesets

rschristian: [We don't use changesets for publishing](https://github.com/preactjs/signals/blob/main/CONTRIBUTING.md), and again, that fix should be a one-off. We really don't want to be supporting `1.2.x` beyond that

XantreGodlike: Removed commit

rschristian:
Correct me if I'm wrong, but looks like this might've already been done?

https://github.com/preactjs/signals/blob/e04671469e9272de356109170b2e429db49db2f0/pnpm-lock.yaml#L1

rschristian:
This looks like it was addressed at some point, so closing it on out

PR #439, opened 2023-11-15
Fix rendering signals as text when using react-transform

andrewiggins:
As suggested by @developit, disable `useSignals` when the auto tracking hook is installed since components can instead rely on that behavior. Also, add `useSignals` to the SignalValue component to fix it when used with the babel transform

### DIFF
--- a/packages/react/runtime/src/auto.ts
+++ b/packages/react/runtime/src/auto.ts
@@ -6,7 +6,7 @@ import {
 import React from "react";
 import jsxRuntime from "react/jsx-runtime";
 import jsxRuntimeDev from "react/jsx-dev-runtime";
-import { EffectStore, useSignals, wrapJsx } from "./index";
+import { EffectStore, wrapJsx, _useSignalsImplementation } from "./index";
 
 export interface ReactDispatcher {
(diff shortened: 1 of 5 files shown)


JoviDeCroock (approved):

packages/react/runtime/src/auto.ts
```diff
@@ -146,11 +146,15 @@ const dispatcherMachinePROD = createMachine({
 ```
 */
 
+export let isAutoSignalTrackingInstalled = false;
+
 let store: EffectStore | null = null;
 let lock = false;
 let currentDispatcher: ReactDispatcher | null = null;
 
 function installCurrentDispatcherHook() {
```
XantreGodlike, line 155: this function still called on top level of `@preact/signals-react` so i there is a difference

andrewiggins: This PR doesn't change the functionality of `@preact/signals-react` in that importing that package still hooks into internals. Changing that behavior will come later since it is a break change. This PR is starting to setup some infra to eventually make that behavior (hooking into React internals) opt-in. So no difference at the moment for people using `@preact/signals-react` but that will change eventually.

PR #443, opened 2023-11-20
Add auto pkg & run react runtimes tests against auto & transform runtimes

andrewiggins:
Add auto package (not exported yet though) in preparation for next major release that will make auto runtime opt-in

Run `@preact/signals-react` test suite against test files using the auto and transform runtimes to verify each runtime's behavior.


packages/react/auto/test/browser/updates.test.tsx
```diff
@@ -0,0 +1,12 @@
+// @ts-ignore-next-line
```
JoviDeCroock, line 1: We could do `ts-expect-error` if it ever resolves

JoviDeCroock (approved):

packages/react/src/index.ts
```diff
@@ -8,8 +8,12 @@ import {
 	untracked,
 } from "@preact/signals-core";
 import type { ReactElement } from "react";
-import { useSignal, useComputed, useSignalEffect } from "../runtime";
-import { installAutoSignalTracking } from "../runtime/src/auto";
```
andrewiggins, line 12: This double import caused some duplicate code in the final bundle. Combining them is what caused this PR to reduce the overall bundle size shown in the size-bot comment.

PR #451, opened 2023-11-27
Removes backward-incompatible type export.

dcporter:
This syntax (export { type Foo }) insta-breaks the build for any project using versions of TS earlier than v4.5. Verified locally that this change fixes the build for earlier versions while not breaking it for later versions.

I believe the only scenario in which this change could impact tree-shaking is if someone is only importing the `ReadonlySignal` type in a project that doesn't otherwise use signals, which seems unlikely.

### DIFF
--- a/.changeset/metal-snakes-hide.md
+++ b/.changeset/metal-snakes-hide.md
@@ -1,0 +1,5 @@
+---
+"@preact/signals-core": patch
+---
+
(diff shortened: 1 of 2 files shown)


dcporter:
The change was originally made by @eddyw in #231 (a very cool PR otherwise) — wanted to @ Eddy in to make sure I'm not missing any big consequences here.

PR #454, opened 2023-11-28
Import runtime package to prevent src import

### LINKED ISSUE #450: signals-react: Wrong import path in dist/signals.d.ts causing TypeScript errors
Hi,

we've discoverd an issue in dist/signals.d.ts that is causing TypeScript errors when building our app. It contains the following line:

```typescript
import { useSignal, useComputed, useSignalEffect } from "../runtime/src/index";
```

That causes the source file `../runtime/src/index` to be included by the TypeScript compiler when building our app causing the TypeScript errors. If I manually change the import path from `src` to `dist` it works fine

paul-sachs:
Using less specific import let's typescript resolve to a d.ts file instead of the .ts file.

Fixes #450

### DIFF
--- a/packages/react/src/index.ts
+++ b/packages/react/src/index.ts
@@ -13,7 +13,7 @@ import {
 	useComputed,
 	useSignalEffect,
 	installAutoSignalTracking,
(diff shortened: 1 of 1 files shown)


paul-sachs:
Closing with preference for #456

Issue #460, opened 2023-12-10
Feature request: atomic transactions


PsychoLlama:
> [!NOTE]
> I built a proof of concept over here: https://github.com/preactjs/signals/pull/461

Hi! First off, let me say I love the library. Beautiful work :clap: 

One thing I'd love to see is an option similar to `batch(...)` for atomic transactions. A common problem with state management is that if a change throws an error halfway through, the state can get corrupted and render the app inoperable. Here's a toy example:

```ts
batch(() => {
  loading.value = true

  // 💥 TypeError: entities.value[index] is undefined
  // Now the app is stuck in a loading state.
  const entityId = entities.value[index].id

  // ... unreachable code ...
})
```

You could imagine this with URL parse errors, out-of-order async actions, timers that fire after the page navigates away... you get the idea. This is even more damaging if you develop in an environment where the user can't refresh.

For most of my app's state transitions, I want these changes to be all-or-nothing. If the update throws an error mid-way through, the entire update is invalid and should be discarded.

```ts
transaction(() => {
  loading.value = true

  // 💥 TypeError: Transaction is aborted.
  const entityId = entities.value[index].id

 // ... unreachable code ...
})

// The in-progress update was aborted, the loading state was rolled back.
loading.value // `false`
```

It's hard to implement this behavior outside the library. I believe it belongs in core.

XantreGodlike:
It's interesting, but I have doubts about this feature. I see no reactivity system which implements it: Vue.js, solid.js, angular. For now I see not a lot of use cases

PsychoLlama:
@XantreGodlike It's not something libraries generally have to implement explicitly, no. It's an implicit property you get from immutable state management systems, particularly Redux. Since every change happens in a "batch" (action) and only committed when the reducer returns, any error in between gets reported without committing intermediate states and leaving the application inconsistent. These systems are widely deployed. With disparate stores like signals, you lose this property.

This isn't a behavior I have to think about when I'm designing applications, but it's a safety net that improves the customer experience. It means more actions can be retried without resorting to refreshes.

For my use case, I'm not using `@preact/signals` directly. I've got a small action harness around it that keeps the benefits of signals (easier code splitting, type safety, value outside a component framework) while preserving the benefits of a Flux architecture (audit log, change observability, free analytics). The only feature I can't implement correctly outside core is atomic rollback.

I can wrap signals with my own getters/setters that restore initial state if an action fails, but after a rollback I can't prevent effects from detecting a version change.

```ts
[...]

PR #461, opened 2023-12-10
Add support for atomic transactions

PsychoLlama:
This is an implementation for my feature request in [issue 460](https://github.com/preactjs/signals/issues/460). It adds a new function called `transaction(...)` which behaves similar to `batch(...)`, except it rolls back all the state changes if an error is thrown.

```ts
const a = signal('original')

try {
  transaction(() => {
    a.value = 'changed'
    throw new Error('kaboom')
  })
} catch {
  a.value // 'original'
}
```

Intermediate states aren't observable until the transaction is complete.

---

I know this change is public facing so I'm not expecting it to be straightforward. Think of the PR as a proof of concept and feel free to say no.


packages/core/src/index.ts
```diff
@@ -96,6 +96,34 @@ function batch<T>(callback: () => T): T {
 	}
 }
 
+// Each nested transaction stores snapshots of the signals it changed.
+const transactionSnapshots: Array<Map<Signal<unknown>, [number, unknown]>> = [];
```
PsychoLlama, line 100: I used a map for simplicity but it seems this library avoids them. I can refactor it to a plain object if there's a preference.

PsychoLlama:
Closing - see [related issue](https://github.com/preactjs/signals/issues/460) for discussion.

PR #462, opened 2023-12-13
Changes void to unknown for callback definitions.

MicahZoltu:
A callback like `() => void` means that the caller must provide a function that is explicitly void returning.  This means one cannot pass a function as a callback that may have a return value that is ignored.  By changing the required return type to `unknown`, it allows the caller to provide either a void returning function, or a function that returns a value.

This is especially valuable when you have callback functions like `() => condition && doThing()`.  Such a function returns a boolean value, which makes it incompatible with `() => void` callback requirements.  You can work around this by wrapping the body with curly braces (to throw away the result of the expression), but it is better to be permissive in what is accepted instead.

### DIFF
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -40,7 +40,7 @@ function hook<T extends OptionsTypes>(hookName: T, hookFn: HookFn<T>) {
 }
 
 let currentComponent: AugmentedComponent | undefined;
-let finishUpdate: (() => void) | undefined;
+let finishUpdate: (() => unknown) | undefined;
 
 function setCurrentUpdater(updater?: Effect) {
 	// end tracking for the current update:
@@ -49,7 +49,7 @@ function setCurrentUpdater(updater?: Effect) {
 	finishUpdate = updater && updater._start();
 }
 
-function createUpdater(update: () => void) {
+function createUpdater(update: () => unknown) {
 	let updater!: Effect;
 	effect(function (this: Effect) {
 		updater = this;
@@ -356,7 +356,7 @@ export function useComputed<T>(compute: () => T) {
 	return useMemo(() => computed<T>(() => $compute.current()), []);
 }
 
-export function useSignalEffect(cb: () => void | (() => void)) {
+export function useSignalEffect(cb: () => unknown | (() => unknown)) {
 	const callback = useRef(cb);
 	callback.current = cb;
 


rschristian (approved):
Sounds good to me -- can this be mirrored into core & react (where applicable)? I imagine our types there will also need an adjustment.

MicahZoltu:
I ran `pnpm changeset`, but no idea if I did so correctly.

Regarding react, I don't use/haven't used react so I'm not sure how this would be ported over to that.

rschristian:
> I ran `pnpm changeset`, but no idea if I did so correctly.

Looks great, thanks! Just helps us keep track of changes and what needs to be released.

> Regarding react, I don't use/haven't used react so I'm not sure how this would be ported over to that.

I might be able to take a look later, but generally it would just be applying the same thing I'd think (+ to `effect()`)? I haven't taken a look at our React adapter in a while though, if it's way more complex then no worries, we can just merge this in.

MicahZoltu:
I do see this, which certainly seems related.  Does that need to be changed as well?
https://github.com/preactjs/signals/blob/main/packages/react/runtime/src/index.ts#L378

packages/preact/src/index.ts
```diff
@@ -356,7 +356,7 @@ export function useComputed<T>(compute: () => T) {
 	return useMemo(() => computed<T>(() => $compute.current()), []);
 }
 
-export function useSignalEffect(cb: () => void | (() => void)) {
+export function useSignalEffect(cb: () => unknown | (() => unknown)) {
```
andrewiggins, line 359: What's the type of the callback passed to `effect`? Should this type match that since this callback is invoked inside of an `effect` and its return value is passed to `effect`? IIRC, `effect` accepts a clean up function

MicahZoltu: Hmm, good point.  We could make the change from
```ts

	useEffect(() => {
		return effect(() => callback.current());
}, []);
```
to
```

	useEffect(() => {
		return effect(() => { callback.current() });
}, []);
```
This would ensure that `effect` is getting a void returning function.  Alternatively, we could change the definition of `effect`, but I think that is in a different repository?

MicahZoltu: I went ahead and made the above change to this PR, can revert if another solution is desired.

PR #467, opened 2023-12-15
Remove auto tracking using React internals from signals-react package

andrewiggins:
Before this change, importing `@preact/signals-react` would invoke side effects that hook into React internals to automatically track signals. This change removes those side effects and requires consumers to update their code to continue using signals in React.

To continue the "auto-tracking" mechanism, users can import `@preact/signals-react/auto` to keep the previous mechanism in place.

I'm also adding support for the comment `@useSignals` and `@noUseSignals` to be consistent with the hook people can use to track signals: `useSignals()`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,10 +4,10 @@
   "scripts": {
     "prebuild": "rimraf packages/core/dist/ packages/preact/dist",
     "build": "pnpm build:core && pnpm build:preact && pnpm build:react-runtime && pnpm build:react-auto && pnpm build:react && pnpm build:react-transform",
-    "_build": "microbundle --raw --globals @preact/signals-core=preactSignalsCore,preact/hooks=preactHooks",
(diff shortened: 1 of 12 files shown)


developit (approved):

package.json
```diff
@@ -4,10 +4,10 @@
   "scripts": {
     "prebuild": "rimraf packages/core/dist/ packages/preact/dist",
     "build": "pnpm build:core && pnpm build:preact && pnpm build:react-runtime && pnpm build:react-auto && pnpm build:react && pnpm build:react-transform",
-    "_build": "microbundle --raw --globals @preact/signals-core=preactSignalsCore,preact/hooks=preactHooks",
+    "_build": "microbundle --raw --globals @preact/signals-core=preactSignalsCore,preact/hooks=preactHooks,@preact/signals-react/runtime=reactSignalsRuntime",
```
rschristian, line 7: Is this global right? I would think it would be `preactSignalsReactRuntime`, per the `@preact/signals-core` -> `preactSignalsCore` earlier

mimiqkz:
Need documentation and examples on this

PR #474, opened 2023-12-22
Re-export useSignals in @preact/signals-react

### LINKED ISSUE #473: @preact/signals-react@2.0.0 not re-exporting useSignals
According [to the docs](https://github.com/preactjs/signals/blob/d7f43ef5c9b6516cd93a12c3f647409cfd8c62be/packages/react/README.md#usesignals-hook), I should be able to `import { useSignals } from '@preact/signals-react'`, but since `useSignals` is not re-exported, I need to do this instead: `import { useSignals } from '@preact/signals-react/runtime'`.

JonAbrams:
Fixes #473

### DIFF
--- a/.changeset/wild-deers-leave.md
+++ b/.changeset/wild-deers-leave.md
@@ -1,0 +1,5 @@
+---
+"@preact/signals-react": minor
+---
+
+Re-export useSignals in @preact/signals-react.
--- a/packages/react/src/index.ts
+++ b/packages/react/src/index.ts
@@ -6,7 +6,8 @@
 // package.json script for this package so their contents aren't bundled into
 // the final source file.
 
-import {
+import type { ReactElement } from "react";
+export {
 	signal,
 	computed,
 	batch,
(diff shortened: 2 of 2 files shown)


.changeset/wild-deers-leave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@preact/signals-react": minor
```
andrewiggins, line 2: Let's make this a `patch` change since this a bug fix with no new feature introductions

andrewiggins (changes requested):
Thanks! Let's adjust the changeset release and get this out :)

andrewiggins:
Actually, just saw that the Readme was updated. Let's roll with that for now. Thanks for opening this though.

rschristian:
@andrewiggins `react-transform` doesn't re-export core though, do we want this?

```js
import { signal, effect } from '@preact/signals-react';
import { useSignals, useSignal } from '@preact/signals-react-runtime';
```

I'm not super in love with this myself.

PR #516, opened 2024-03-06
chore: remove unnecessary try...finally

OnlyWick:
there are no operations in the `notify` method that could cause program interruption.


OnlyWick:
Will it happen when used together with other frameworks? I see in the `core` that `_notify` does not throw any errors.

PR #533, opened 2024-03-18
build(actions): update actions, use actions/setup-node's builtin caching

jviide:
This pull request updates most actions used in the CI workflows, and makes the following changes:
 * Use actions/setup-node's builtin dependency caching for pnpm as described [here](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-data).
 * Related to the change described above, run pnpm/action-setup before actions/setup-node.
 * Remove the `run_install: false` option from pnpm/action-setup, as it's the default.

### DIFF
--- a/.github/workflows/compressed-size.yml
+++ b/.github/workflows/compressed-size.yml
@@ -1,4 +1,5 @@
 name: compressed-size
+
 on:
   pull_request:
     branches:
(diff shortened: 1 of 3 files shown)


marvinhagemeister (approved):
Sweet 💯

PR #543, opened 2024-03-27
chore(hooks): removed usage of useMemo instead using useConst implemented with useRef

danilocastro-shopify:
Proposing to prevent using hooks based on array dependency to simplify data lifecycle.

- It does not create multiple empty arrays for future GC collection
- Simple check  to create the instance

Maybe this change is small and not worthy. Considering the usage of several hooks around the application, it would reduce the GC time slices.

### DIFF
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -33,18 +33,14 @@ const HAS_PENDING_UPDATE = 1 << 0;
 const HAS_HOOK_STATE = 1 << 1;
 const HAS_COMPUTEDS = 1 << 2;
 
+const UNINITIALIZED = {};
 
 function useConst<T>(fn: () => T): T {
-	interface RefManagement {
-		value: T
-		status: 'uninitialized' | 'initialized'
+	const ref = useRef<typeof UNINITIALIZED | T>(UNINITIALIZED);
+	if (ref.current === UNINITIALIZED) {
+		ref.current = fn();
 	}
-	const ref = useRef<RefManagement>({status: 'uninitialized'} as any)
-	if (ref.current.status === 'uninitialized') {
-		ref.current = {status: 'initialized', value: fn()}
-	}
-
-	return ref.current.value
+	return ref.current;
 }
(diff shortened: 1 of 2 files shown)


packages/preact/src/index.ts
```diff
@@ -33,6 +33,21 @@ const HAS_PENDING_UPDATE = 1 << 0;
...
+		value: T
+		status: 'uninitialized' | 'initialized'
+	}
+	const ref = useRef<RefManagement>({status: 'uninitialized'} as any)
+	if (ref.current.status === 'uninitialized') {
+		ref.current = {status: 'initialized', value: fn()}
+	}
+
+	return ref.current.value
+}
```
developit, line 48:
```suggestion
const UNINITIALIZED = {};

function useConst<T>(fn: () => T): T {
	const ref = useRef<typeof UNINITIALIZED | T>(UNINITIALIZED);
	if (ref.current === UNINITIALIZED) {
		ref.current = fn();
	}
	return ref.current;
}
```

packages/react/runtime/src/index.ts
```diff
@@ -360,19 +360,32 @@ Object.defineProperties(Signal.prototype, {
...
+		value: T
+		status: 'uninitialized' | 'initialized'
+	}
+	const ref = useRef<RefManagement>({status: 'uninitialized'} as any)
+	if (ref.current.status === 'uninitialized') {
+		ref.current = {status: 'initialized', value: fn()}
+	}
+
+	return ref.current.value
+}
```
developit, line 374:
```suggestion
const UNINITIALIZED = {};

function useConst<T>(fn: () => T): T {
	const ref = useRef<typeof UNINITIALIZED | T>(UNINITIALIZED);
	if (ref.current === UNINITIALIZED) {
		ref.current = fn();
	}
	return ref.current;
}
```

JoviDeCroock (approved):
Sorry for getting so late to this, can we add a changeset and rebase this branch?

Issue #554, opened 2024-04-05
[Contribution]: Build is not works on windows and inside Github codespaces


XantreDev:
`mv`, `rm -dr` - from scripts is not valid command on windows. [reference](https://github.com/preactjs/signals/blob/d3935e867646ba2628853c236c9ec7efdfc7aaee/package.json#L16)
`rm -dr` is not valid in Github codespaces either (only some of the rm flags supported in VM)

Can I make scripts cross platform to make contributions easier?

rschristian:
Sure, we're already using `rimraf` (inconsistently) anyhow (such as the `prebuild`)

rschristian:
Addressed in #554

PR #555, opened 2024-04-06
Fixed build on windows

XantreDev:
#554


package.json
```diff
@@ -89,6 +89,7 @@
     "mocha": "^10.0.0",
     "prettier": "^2.7.1",
     "rimraf": "^3.0.2",
+    "shx": "^0.3.4",
```
rschristian, line 92: As we just need `mv`, I think we can find a simpler/smaller package for that.

I can't recall the name but I'm fairly certain there's a tiny (Windows-compatible) `mv` CLI package, sorta like `rimraf`.

XantreDev: When I'm done with research, I found only this one. 
I can write a script in JavaScript if you prefer it

rschristian: [`move-cli`](https://npm.im/move-cli) perhaps? I guess alternatively we could drop `rimraf` for `shx`?

XantreDev: I've replaced rimraf with `shx`

rschristian (approved):

PR #558, opened 2024-04-13
fix(core): Restore stricter effect callback & cleanup function types

jviide:
This pull request restores the stricter types for effect callbacks and effect cleanup functions introduced in #249. Core tests needed to be modified slightly to accommodate these typing changes.

These stricter types were later changed back to allowing both effect callbacks and effect cleanup functions return `unknown`. It is convenient to allow anything to be returned, especially when writing tests like `effect(() => s.value++)`. On the other hand stricter types can be useful for catching e.g. attempts to have asynchronous effect handlers, which the core does not support (`effect(async () => { ... })`).

It's debatable whether this change would warrant a patch or major release. I've tentatively chosen a patch changeset, as the previous related changes were also patch changes.

### DIFF
--- a/.changeset/nine-fishes-leave.md
+++ b/.changeset/nine-fishes-leave.md
@@ -1,0 +1,5 @@
+---
+"@preact/signals-core": patch
+---
+
+Restore stricter effect callback & cleanup function types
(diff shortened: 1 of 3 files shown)


marvinhagemeister (approved):
Awesome, thanks for improving this 💯

PR #570, opened 2024-06-12
Fix out-of-order effect error when suspending in React Native

developit:
I have not managed to reproduce this in tests, since it only seems to happen in React Native, and I believe only with Hermes. In this environment, layout effects are flushed after a microtask (or synchronously?), and rendering can be triggered synchronously from them (via setState/useSyncExternalStore/etc). This means the "close current tracking context" operation we schedule after a Promise tick doesn't happen before the next render begins, which can cause an unmanaged useSignals() call to be left "open".

I've added a `useLayoutEffect()` to unmanaged `useSignals()` to address this, which just works by closing any open tracking context when React finishes the current render (as defined by the time at which React begins to execute effects).

As part of this, I also had to tweak `store.f()` to ignore repeated calls, as doing so could also trigger the "Out-of-order effect" exception (`endEffect()` does not guard against this internally as its only possible to trigger when reaching into the guts of signals-core as we do in the integration libraries).

### DIFF
--- a/packages/react/runtime/test/browser/suspense.test.tsx
+++ b/packages/react/runtime/test/browser/suspense.test.tsx
@@ -111,7 +111,12 @@ describe("Suspense", () => {
 		}
 
 		await render(<Parent />);
-		expect(scratch.innerHTML).to.equal("<span>loading...</span>");
+		expect(scratch.innerHTML).to.be.oneOf([
+			// react 17+
+			`<span>loading...</span>`,
+			// react 16
+			`<p style="display: none !important;">0</p><span>loading...</span>`,
+		]);
 
 		await act(async () => {
 			signal1.value++;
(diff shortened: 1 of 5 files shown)


packages/react/runtime/test/browser/suspense.test.tsx
```diff
@@ -0,0 +1,146 @@
...
+} from "@preact/signals-react/runtime";
+import {
+	Root,
+	createRoot,
+	act,
+	checkHangingAct,
+	getConsoleErrorSpy,
+} from "../../../test/shared/utils";
+
+describe.only("Suspense", () => {
```
JoviDeCroock, line 16: Remaining only

JoviDeCroock (approved):

packages/react/runtime/test/browser/suspense.test.tsx
```diff
@@ -0,0 +1,149 @@
...
+					<Child />
+					<Middle>
+						<SuspendingComponent />
+					</Middle>
+				</Suspense>
+			);
+		}
+
+		await render(<Parent />);
+		expect(scratch.innerHTML).to.equal("<span>loading...</span>");
```
JoviDeCroock, line 114: These assertions do not take the default React view in account, i.e. when React suspends we will see `<p style="display:none;">{signal1.value}</p><your-placeholder>`

PR #577, opened 2024-06-25
chore(signals-react): throw error when auto-tracking with React 19

### LINKED ISSUE #576: installAutoSignalTracking fails with React 19
- [X] Check if updating to the latest version resolves the issue

**Environment**
- [X] I am using `@preact/signals-react`
  - [X] I understand usage changed in v2, and I've followed the [React Integration instructions](https://github.com/preactjs/signals/tree/chore/issue-templates/packages/react#react-integration)

**Describe the bug**
If you call
```
import { installAutoSignalTracking } from '@preact/signals-react/runtime';
installAutoSignalTracking();
```

when using
```
react@19.0.0-rc-3563387fe3-20240621
react-dom@19.0.0-rc-3563387fe3-20240621
```

it fails with
```
TypeError: Cannot read properties of undefined (reading 'ReactCurrentDispatcher')
```

in addition to complaining about the peer dependency (`react@"^16.14.0 || 17.x || 18.x"`) not including 19.x

**To Reproduce**
[Please provide a link to a StackBlitz/CodeSandbox/Codepen project or a GitHub repository that demonstrates the issue.
](https://github.com/Artur-/react19-signals-tests)

1. Open the app

**Expected behavior**
No error

JoviDeCroock:
Resolves #576 

This should point people in the right direction 😅

### DIFF
--- a/packages/react/runtime/src/auto.ts
+++ b/packages/react/runtime/src/auto.ts
@@ -2,6 +2,7 @@ import {
 	// @ts-ignore-next-line
 	// eslint-disable-next-line @typescript-eslint/no-unused-vars
 	__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED as ReactInternals,
+	version as reactVersion,
 } from "react";
 import React from "react";
 import jsxRuntime from "react/jsx-runtime";
@@ -378,6 +379,12 @@ export function installJSXHooks() {
 }
 
 export function installAutoSignalTracking() {
+	const [major] = reactVersion.split(".").map(Number);
+	if (major >= 19) {
+		throw new Error(
+			"Automatic signals tracking is not supported in React 19 and later, try the Babel plugin instead https://github.com/preactjs/signals/tree/main/packages/react-transform#signals-react-transform."
+		);
+	}
 	installCurrentDispatcherHook();
 	installJSXHooks();
 }
--- a/packages/react/runtime/src/auto.ts
+++ b/packages/react/runtime/src/auto.ts
@@ -382,7 +382,7 @@ export function installAutoSignalTracking() {
 	const [major] = reactVersion.split(".").map(Number);
 	if (major >= 19) {
 		throw new Error(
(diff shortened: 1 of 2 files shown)


rschristian (approved):
Is the peer dependency warning not enough? I sorta figure that should be enough, but I guess this can't hurt

marvinhagemeister (approved):

packages/react/runtime/src/auto.ts
```diff
@@ -378,6 +379,12 @@ export function installJSXHooks() {
 }
 
 export function installAutoSignalTracking() {
+	const [major] = reactVersion.split(".").map(Number);
+	if (major >= 19) {
+		throw new Error(
+			"Automatic signals tracking is not supported in React 19 and later, try the Babel plugin instead https://github.com/preactjs/signals/tree/main/packages/react-transform#signals-react-transform."
```
developit, line 385: Might be good to shorten the URL.

JoviDeCroock:
```suggestion
			"Automatic signals tracking is not supported in React 19 and later, try the Babel plugin instead https://bit.ly/3zbndcL."
```

rschristian: Bit late, but I think most would prefer a bare & long URL over a shortened link leading god only knows where -- link shorteners tend to scream "sketchy".

JoviDeCroock: 😅 thought it was just me, I'll revert it

packages/react/runtime/src/auto.ts
```diff
@@ -378,6 +379,12 @@ export function installJSXHooks() {
 }
 
 export function installAutoSignalTracking() {
+	const [major] = reactVersion.split(".").map(Number);
+	if (major >= 19) {
+		throw new Error(
+			"Automatic signals tracking is not supported in React 19 and later, try the Babel plugin instead https://bit.ly/3zbndcL."
```
JoviDeCroock, line 385:
```suggestion
			"Automatic signals tracking is not supported in React 19 and later, try the Babel plugin instead https://github.com/preactjs/signals/tree/main/packages/react-transform#signals-react-transform."
```

Issue #532, opened 2024-03-18
The parameter of `singal()` can be omitted, just like `useState()`. 


cnbebop:
Think about this, create a issue signal, and it's initial is undefined, then you should pass `undefined` to both generic parameter and function parameter.

```ts
const issue = signal<Issue | undefined>(undefined);
```

We can see that it leads to boilerplate code and the development experience is reduced.

The `signal()` can be defiened like this:

```ts
function signal<T>(value: T): Signal<T>;
function signal<T = undefined>(): Signal<T | undefined>;
```

XantreGodlike:
I don't think it's a problem. Better to keep function monomorphic with the same amount of arguments

cnbebop:
> I don't think it's a problem. Better to keep function with the same amount of arguments (it's better for performance)

Even if you're right, but the class `Signal` has a optional parameter contructor, how to explain it?

rschristian:
Looks reasonable to me, if you want to PR it.

XantreGodlike:
> > I don't think it's a problem. Better to keep function with the same amount of arguments (it's better for performance)
> 
> Even if you're right, but the class `Signal` has an optional parameter constructor, how to explain it?
It's not for public usage. But I see, probably it will think that it can have 0 params
https://github.com/preactjs/signals/blob/58befba577d02c5cac5292fda0a599f9708e908b/packages/core/src/index.ts#L531

Anyway I think it should be benchmarked. Check for deoptimizations and how much it slows other benchmarks

rschristian:
> Anyway I think it should be benchmarked. Check for deoptimizations and how much it slows other benchmarks

This should be a type signature-only change. There will be no effect on the benchmarks

JoviDeCroock:
@XantreGodlike This is in relation to an already allowed paradigm, signals can be initiated with undefined or a missing initializer as you yourself link to for the `Computed.prototype`. This can't possibly affect benchmarks as it's merely a change in how the generic infers the starting value.

XantreGodlike:
It's not we are starting to provide interface that allows different amount of args. It can cause performance slow down, depending on usage
https://jsbench.me/7slty3mtij/1
<img width="715" alt="image" src="https://github.com/preactjs/signals/assets/57757211/d498d705-77a5-441a-a066-a936aaf07760">

XantreGodlike:
But probably impact will be insignificant

JoviDeCroock:
[...]

PR #578, opened 2024-06-25
chore(signals): add default undefined type

JoviDeCroock:
Resolves https://github.com/preactjs/signals/issues/532

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -392,7 +392,9 @@ Object.defineProperty(Signal.prototype, "value", {
  * @param value The initial value for the signal.
  * @returns A new signal.
  */
-function signal<T>(value: T): Signal<T> {
+export function signal<T>(value: T): Signal<T>;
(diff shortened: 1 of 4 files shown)


packages/core/src/index.ts
```diff
@@ -821,4 +823,4 @@ function effect(fn: EffectFn): () => void {
 	return effect._dispose.bind(effect);
 }
 
-export { signal, computed, effect, batch, untracked, Signal, ReadonlySignal };
+export { computed, effect, batch, untracked, Signal, ReadonlySignal };
```
rschristian, line 826: Is this intended?

rschristian (approved):

packages/core/src/index.ts
JoviDeCroock: Yes, overloads need to be exported alongside their definition 😅

rschristian: Code blindness, didn't even notice the `export` statements above. Sorry!

Issue #552, opened 2024-04-04
Components using Array.map to construct children do not satisfy conditions for the transform


alxclark:
Using a simple component that maps onto an array, the transform does not see that the children are actually using signals and does not trigger the transform.

**Example:**

```tsx
export function App({ name }) {
  const greetings = ["Hello", "Goodbye"]
  const children = greetings.map((greeting) => <div key={greeting}>{greeting} {name.value}</div>)

  return <div>{children}</div>;
}
```

I opened a draft PR adding a test showing the error case: https://github.com/preactjs/signals/pull/551/files#r1552520125

alxclark:
After digging into the babel plugin for a little while, it looks like the following happens when using a map:

* We iterate over all `MemberExpression`
* We find the signal access in `Array.map`
* We attempt to set on the parent function scope that this function uses signals
* We end up writing on the arrow function of the map instead of the component
* We iterate over function calls, check if they both contain JSX and signals
* Turns out the function is not marked as using signals since it was added to the map's arrow function scope.
* Transform ignores our component.

A naive and not performant solution would be to recursively write to the parent until we hit the root of our program, which should hopefully be just a handful of levels deep.

There's probably a better way of solving this, I'd love some feedback from folks more familiar with babel plugins. 😄

XantreDev:
Yes, we can do it here, until reach parent
https://github.com/preactjs/signals/blob/d3935e867646ba2628853c236c9ec7efdfc7aaee/packages/react-transform/src/index.ts#L541

XantreDev:
I can make a pr. It shouldn't impact performance

alxclark:
@XantreDev That would be awesome, I'm more than happy to review once you do!

PR #553, opened 2024-04-05
Addes recursive `.value` read detection in nested functions

XantreDev:
#552

PR #582, opened 2024-07-04
fix(react-signals-transform): Support nested functions accessing signals

JoviDeCroock:
This was a bit more complex than initially anticipated due to how we have chosen to transform, I wanted to make as little change as possible so here goes...

- This adds a check to our `memberExpression` that zooms out towards the component definition
- If along the way we encounter any hooks like `useEffect`/... we bail
- When we reach the highest scope we signal this as a component in need of signals

Fixes https://github.com/preactjs/signals/issues/552
Supersedes https://github.com/preactjs/signals/pull/553

### DIFF
--- a/packages/react-transform/src/index.ts
+++ b/packages/react-transform/src/index.ts
@@ -63,7 +63,6 @@ function getComponentFunctionDeclaration(
 		const parent = functionScope.path.parent;
 		let functionName = getFunctionName(functionScope.path as any);
 		if (functionName === DefaultExportSymbol) {
-			// TODO
 			functionName = filename || null;
 		}
 		if (isComponentFunction(functionScope.path as any, functionName)) {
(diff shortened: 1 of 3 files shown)


JoviDeCroock:
Apparently I missed the Suspense tests, I would probably also need to look at higher order components before feeling confident as we might need an early return case for hoc's

packages/react-transform/src/index.ts
```diff
@@ -52,8 +52,47 @@ const setData = (node: DataContainer, name: string, value: any) =>
...
+	filename: string | undefined,
+	prev?: Scope
+): Scope | null {
 	const functionScope = path.scope.getFunctionParent();
+
+	if (functionScope) {
+		const parent = functionScope.path.parent;
+		let functionName = getFunctionName(functionScope.path as any);
+		if (functionName === DefaultExportSymbol) {
+			// TODO
```
JoviDeCroock, line 66:
```suggestion
```

rschristian (approved):

Issue #575, opened 2024-06-24
@preact/signals-react-transform not working in React Native


phanthaoIT:
- [ ] Check if updating to the latest version resolves the issue

**Environment**
- [ ] I am using `@preact/signals-core`
- [ ] I am using `@preact/signals`
- [X] I am using `@preact/signals-react`
  - [X] I understand usage changed in v2, and I've followed the [React Integration instructions](https://github.com/preactjs/signals/tree/chore/issue-templates/packages/react#react-integration)

**Describe the bug**
I added babel plugin but it doesn't work in React Native. it throws error: ```Cannot import useSignals outside of an ESM module file```
<img width="655" alt="image" src="https://github.com/preactjs/signals/assets/31787799/c23f7df0-be58-48de-ad07-0bfca3ed8dde">
<img width="786" alt="image" src="https://github.com/preactjs/signals/assets/31787799/d347d783-4ac5-4a26-905a-0d6d2a471f7a">

rschristian:
Well the error is correct, you cannot use `import` outside of an ESM context.

Our export map is correct & providing the right formats, do you mean to use `require` instead?

phanthaoIT:
Hi @rschristian, I followed the guide to add Babel transform, My project is just a React Native app but the babel  has the error```Cannot import useSignals outside of an ESM module file``` .  The signal can work but I need to add manual ```import { useSignals } from '@preact/signals-react/runtime';```. it should be works automatic

rschristian:
> My project is just a React Native app 

We need an actual reproduction in order to help, this isn't enough to go on.

We do ship both [CJS & ESM builds](https://github.com/preactjs/signals/blob/931404e96338e120464b73e522148389e38eeb2b/packages/react-transform/package.json#L29-L35), your config may be pulling the wrong one?

phanthaoIT:
Hi @rschristian, I think the issue is in my repo, I compile everything to CommonJS. Do you know how to signal works for this case?

JoviDeCroock:
I reckon because you preface the babel import with `module:` it will default to import syntax, no?

XantreDev:
I provide support of react-native in @preact-signals/safe-react
I would like to contribute this feature to the mainstream library when I will have time

phanthaoIT:
I appreciate your support! But after I upgrade the latest version of @preact/signals-react-transform, I got the error  
[...]

PR #584, opened 2024-07-12
feat(react-transform): Suppport require syntax in our babel transform

JoviDeCroock:
Noticed we threw an error for non-ESM modules, so added support for that as we are exporting a CJS build of our react-bindings either way.

CC @andrewiggins not sure if you opted out here for a reason

Fixes https://github.com/preactjs/signals/issues/575

### DIFF
--- a/.changeset/wet-radios-clap.md
+++ b/.changeset/wet-radios-clap.md
@@ -1,0 +1,5 @@
+---
+"@preact/signals-react-transform": minor
+---
+
+Support `require()` syntax in the Babel transform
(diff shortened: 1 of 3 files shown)


marvinhagemeister (approved):

marvinhagemeister:
LGTM, but a test case would be nice

JoviDeCroock:
@marvinhagemeister good shout, was still figuring out how to trick the `isModule` helper 😅

rschristian (approved):
LGTM, other than the thing I mentioned in Slack.

I imagine this doesn't need any altered instructions for the end user then? They can add the Babel plugin in the same way and all will work?

alxrodav:
Any progress with this PR? I started a new React Native App, and would love to give it a try to signals

Issue #585, opened 2024-07-19
`ReadonlySignal` should not inherit from `Signal` type


KillariDev:
**Environment**
- [x] I am using `@preact/signals`

**Describe the bug**

**To Reproduce**
```ts
const numberSingal: ReadonlySignal<number> = useSignal<number>(0) // ok
const numberSignal: Signal<number> = useComputed<number>(0) // should error
```

This assignment should not be allowed as signal is not readonly

the bug is here:
https://github.com/preactjs/signals/blob/main/packages/core/src/index.ts#L660
```ts
interface ReadonlySignal<T = any> extends Signal<T> {
	readonly value: T;
}
```

**Expected behavior**
There should be a type error

rschristian:
This is the intended behavior I believe.

The purpose of `ReadonlySignal` is to stop you from writing to an incoming signal, not assert whether the signal can be written to. For example, a component might take a `ReadonlySignal` to show it only consumes, it won't try to write to that data.

KillariDev:
> This is the intended behavior I believe.
> 
> The purpose of `ReadonlySignal` is to stop you from writing to an incoming signal, not assert whether the signal can be written to. For example, a component might take a `ReadonlySignal` to show it only consumes, it won't try to write to that data.
> 
> Edit: Besides, the only API to return `ReadonlySignal` is `computed`/`useComputed`, and it makes no sense for a consumer to distinguish between them.
> 
> As such, I'll close this on out.

I had a bug in my code where I had a component that consumed a `ReadonlySignal`, but did not edit it. I then computed an `ReadonlySignal` and passed to it. Everything good. Then I modified the original component to edit the passed on signal and made the component `Signal`. All was allegedly good in compiling part, but then when I run the code, I got runtime exception which should have been caught at compilation time.

This can be fixed by doing the inheritance the other way around. This is clearly a bug in the library as the library is using typescript wrong.

this is like `"Apple"` is a string, that should be able to things that expect string. But If a function expects `"Apple"`, passing a string to it is not sufficient, as the string could be `"Banana"`, which clearly is not `"Apple"`, while both of them are strings.

"Apple" is of type string, but string is not type of "Apple"
"Banana" is of type string, but string is not type of "Banana"

rschristian:
> All was allegedly good in compiling part, but then when I run the code, I got runtime exception which should have been caught at compilation time.

[...]

PR #587, opened 2024-08-02
Adjust ReadOnlySignal so we cant widen the type

JoviDeCroock:
Avoid inheriting from signal in ReadonlySignal so we can't accidentally widen the type by doing `const computed: Signal<any> = useComputed(() => {})`, in widening this type we would lose the `readonly` aspect of the value property effectively making it a mutable signal.

Fixes https://github.com/preactjs/signals/issues/585

### DIFF
--- a/.changeset/seven-seahorses-cry.md
+++ b/.changeset/seven-seahorses-cry.md
@@ -1,6 +1,16 @@
 ---
-"@preact/signals-core": patch
+"@preact/signals-core": minor
 ---
 
-Adjust ReadOnlySignal to not inherit from `Signal`
-so the type can't be widened
+Adjust the `ReadOnlySignal` type to not inherit from `Signal`
+this way the type can't be widened without noticing, i.e. when
+we'd have
+
+```js
+const sig: Signal = useComputed(() => x);
+```
+
+We would have widened the type to be mutable again, which for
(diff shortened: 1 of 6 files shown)


.changeset/seven-seahorses-cry.md
```diff
@@ -0,0 +1,6 @@
+---
+"@preact/signals-core": patch
```
JoviDeCroock, line 2: This might need to be a major though 😅

.changeset/seven-seahorses-cry.md
```diff
@@ -0,0 +1,6 @@
+---
+"@preact/signals-core": minor
```
JoviDeCroock, line 2: This might need to be a major though 😅

rschristian: Minor for a type correction is completely a-ok, IMO.

rschristian (approved):

.changeset/seven-seahorses-cry.md
JoviDeCroock: I'll expand on the changeset then to provide our reasoning to avoid excessive issues about it

marvinhagemeister (approved):

Issue #315, opened 2023-03-04
Timing issue with useComputed


jakearchibald:
https://codesandbox.io/embed/charming-rosalind-qyxbh1?file=/src/index.js&codemirror=1

Here, a component is only rendered if the value of a signal is not-null.

Within that component, a `useComputed` assumes the signal is not-null. However, the recomputation appears to run before the component is unmounted, meaning it receives a null value.

JoviDeCroock:
Hmm yes, this seems to be a side-effect of mixing a synchronous update with an asynchronous one. When we can't do an optimal update (like synchronously updating text or an attribute) we have to fall back to re-rendering the VNode, which in this case leads to changing the return value from the VNode to that dom-node.

However the dependencies does include one that we can do synchronously i.e. that recompute which will happen synchronously with the null value while the VNode update will be scheduled like any other update in Preact, which would allow us to batch it with useState/....

jakearchibald:
> I guess it could be more intuitive if we say that computeds used as hooks are hooked into the Preact render cycle

I think that's probably the safest thing. I ran into another bug that might also be related to timing. I'll file it today.

jviide:
How about something like this? Sorry, wrote this in JavaScript instead of TS so I could test it quickly in the sandbox:

```ts
function useComputed(compute) {
  // Keep track of the latest compute function.
  const $func = useRef(compute);
  $func.current = compute;

  // Keep track of the latest compute function result
  // so that we can fall back to it after unmount.
  const $value = useRef();

  // On unmount just start returning the last computed value.
  useEffect(() => {
    $func.current = null;
  }, []);

  const $compute = useRef();
  if (!$compute.current) {
    $compute.current = computed(() => {
      if ($func.current) {
        $value.current = $func.current();
      }
      return $value.current;
    });
  }

  // Update the internal flag here?
  // (currentComponent as AugmentedComponent)._updateFlags |= HAS_COMPUTEDS;
  return $compute.current;
}
```

This way the computed's updates end in tandem with the component's lifecycle, and it'll just start returning the last computed value whenever asked.

PR #595, opened 2024-08-25
RAF aligned effects

JoviDeCroock:
At first I thought this would solve https://github.com/preactjs/signals/issues/315 but that's still not the case I think as we are currently only deferring explicit effects, however in this case it's the distinction between re-rendering a VNode and an eagerly recalculating computed.

I guess in an ideal scenario we don't want to allow for synchronous updates within a Preact application, doing so will force us into scenario's where our DOM-tree is newer than our VNode tree

#315 is a perfect example of that going wrong 😅 however this PR will be a performance improvement nonetheless as we are not blocking anymore.

I would consider finding a solution to #315 important for mixed usage between DOM-optimised updates and conditional VDOM.

### DIFF
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -377,7 +377,9 @@ function notify(this: Effect) {
 	queue.push(this);
 	if (!isFlushing) {
 		isFlushing = true;
-		(typeof window === "undefined" ? setTimeout : requestAnimationFrame)(flush);
+		(typeof requestAnimationFrame === "undefined"
+			? setTimeout
+			: requestAnimationFrame)(flush);
 	}
 }
 
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -377,11 +377,7 @@ function notify(this: Effect) {
 	queue.push(this);
 	if (!isFlushing) {
 		isFlushing = true;
-		(process.env.TEST
-			? (cb: () => void) => cb()
-			: typeof window === "undefined"
-			? setTimeout
-			: requestAnimationFrame)(flush);
+		(typeof window === "undefined" ? setTimeout : requestAnimationFrame)(flush);
 	}
 }
 
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -359,21 +359,25 @@ export function useComputed<T>(compute: () => T) {
 }
 
 let oldNotify: (this: Effect) => void,
-	queue: Array<Effect> = [];
+	queue: Array<Effect> = [],
+	isFlushing = false;
 
 function flush() {
 	batch(() => {
-		let inst: Effect | undefined;
-		while ((inst = queue.shift())) oldNotify.call(inst);
+		let flushing = [...queue];
+		isFlushing = false;
+		queue.length = 0;
+		for (let i = 0; i < flushing.length; i++) {
+			oldNotify.call(flushing[i]);
+		}
 	});
 }
 
 function notify(this: Effect) {
 	queue.push(this);
-	if (queue.push(this) === 1) {
-		(typeof requestAnimationFrame === "undefined"
-			? setTimeout
-			: requestAnimationFrame)(flush);
+	if (!isFlushing) {
+		isFlushing = true;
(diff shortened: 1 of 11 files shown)


JoviDeCroock:
Hah odd, React in production has different values than dev...

packages/preact/src/index.ts
```diff
@@ -358,12 +358,43 @@ export function useComputed<T>(compute: () => T) {
...
+			oldNotify.call(flushing[i]);
+		}
+	});
+}
+
+function notify(this: Effect) {
+	queue.push(this);
+	if (!isFlushing) {
+		isFlushing = true;
+		(process.env.TEST
```
JoviDeCroock, line 380: Goal is to remove this and use an afterFrame helper in the tests when this is in the finalization phase

packages/preact/src/index.ts
```diff
@@ -358,12 +358,39 @@ export function useComputed<T>(compute: () => T) {
...
+			oldNotify.call(flushing[i]);
+		}
+	});
+}
+
+function notify(this: Effect) {
+	queue.push(this);
+	if (!isFlushing) {
+		isFlushing = true;
+		(typeof window === "undefined" ? setTimeout : requestAnimationFrame)(flush);
```
XantreDev, line 380: Should it fallback to `queueMicrotask` to preserve old behavior in environments that doesn't have window object? (React Native)

JoviDeCroock: Window should be defined in React native and so should `requestAnimationFrame` iirc, RN does something like `global.window = global`

marvinhagemeister (approved):
LGTM, apologies for the late review. I missed it in my notifications

packages/preact/src/index.ts
XantreDev: I am pretty much sure there is no window. I will recheck. But there is requestAnimationFrame for sure

developit, line 380:
```suggestion
		(typeof requestAnimationFrame === 'undefined' ? setTimeout : requestAnimationFrame)(flush);
```

packages/preact/src/index.ts
```diff
@@ -358,12 +358,39 @@ export function useComputed<T>(compute: () => T) {
...
+	isFlushing = false;
+
+function flush() {
+	batch(() => {
+		let flushing = [...queue];
+		isFlushing = false;
+		queue.length = 0;
+		for (let i = 0; i < flushing.length; i++) {
+			oldNotify.call(flushing[i]);
+		}
```
developit, line 372:
```suggestion
		let inst;
		while ((inst = queue.shift()) oldNotify.call(inst);
```

packages/preact/src/index.ts
```diff
@@ -358,12 +358,39 @@ export function useComputed<T>(compute: () => T) {
 	return useMemo(() => computed<T>(() => $compute.current()), []);
 }
 
+let oldNotify: (this: Effect) => void,
+	queue: Array<Effect> = [],
+	isFlushing = false;
```
developit, line 363:
```suggestion
	queue: Array<Effect> = [];
```

packages/preact/src/index.ts
```diff
@@ -358,12 +358,39 @@ export function useComputed<T>(compute: () => T) {
...
+		for (let i = 0; i < flushing.length; i++) {
+			oldNotify.call(flushing[i]);
+		}
+	});
+}
+
+function notify(this: Effect) {
+	queue.push(this);
+	if (!isFlushing) {
+		isFlushing = true;
```
developit, line 379:
```suggestion
	if (queue.push(this) === 1) {
```

developit:
We could also apply the same notification delay to [the prop updater](https://github.com/preactjs/signals/blob/b00e64dd448f6f47d9db3635c5c8304c9e436543/packages/preact/src/index.ts#L241) and [the text updater](https://github.com/preactjs/signals/blob/b00e64dd448f6f47d9db3635c5c8304c9e436543/packages/preact/src/index.ts#L99) to limit their DOM mutation frequency to screen refresh.

JoviDeCroock:
@developit I'll follow up the changes to props/text separately

PR #600, opened 2024-09-21
chore: Support running a subset of tests

rschristian:
It's a tad annoying to work with the test suite as there's effectively no way to focus tests -- at a minimum, [this will throw](https://github.com/preactjs/signals/blob/e04671469e9272de356109170b2e429db49db2f0/karma.conf.js#L34) if you haven't built the React transform (which is quite commonly the case for me as I have no reason to touch the React integration).

While not quite Karma's `--grep` flag, this supports `--grep=core,preact,...` to optionally filter & focus the test suite.


package.json
```diff
@@ -45,6 +45,7 @@
   "license": "MIT",
   "devDependencies": {
     "@babel/core": "^7.23.3",
+    "@babel/helper-module-imports": "^7.22.5",
```
rschristian, line 48: The three deps moved/copied to the root are to support this: https://github.com/preactjs/signals/blob/e04671469e9272de356109170b2e429db49db2f0/test/browser/babel.js#L10

I'm not sure if that can support being set async w/out race conditions, so I just move the deps up. If they weren't, the tests can still succeed, but there will be a bunch of rather noisy errors printed to the log.

rschristian:
I have no idea what PNPM's problem is, other than that after upgrading lockfile version, it seems to have picked entirely new versions of at least iso & some babel dep, ignoring previous locked versions. Will try to upgrade that separately.

package.json
rschristian: Turns out these setup functions are only needed for the React tests. Whoops! Can be filtered out entirely if no react suites are being ran.

JoviDeCroock (approved):

PR #604, opened 2024-10-03
Raf aligned dom updates

JoviDeCroock:
This also converts our synchronous DOM updates to be animation-frame aligned, this will make them execute in parallel and when the browser is prepared for it.

Let's draw out the ideal scenario we want to create:

- effects that end up in a setState get executed synchronously so these updates are immediately added in the render queue
- signal-effects are deferred by an animation-frame
- TextNodeUpdates are deferred by an animation-frame
- AttributeNodeUpdates are deferred by an animation-frame

Now we can be sure that Preact updates take presedence over our normal DOM updates.

One thing I am unsure of is the following https://github.com/preactjs/signals/blob/main/packages/preact/src/index.ts#L92-L100 whether we can reliably make this also deferred as currently it's synchronous, we could just defer the `s.data = s.peek()` part? From the tests it looks to work as we expect with RAF though 😅 

Fixes https://github.com/preactjs/signals/issues/315
Proof: https://codesandbox.io/p/sandbox/epic-wilbur-hhln8l

This could be considered a breaking change though 😅 

Another thing to consider here would be the testing story, we could tie it into `options.debounceRendering` but that might lead to some odd behavior in terms of rendering. To align with how the user would see this behave it would need to execute Preact updates and then update the nodes directly.

In the latest version of this branch the above is fixed by leveraging `options.requestAnimationFrame` which now means that any signal update needs to be wrapped in `act`. In a way this was always the case as a signal update that resulted in a Preact state update would require it to be wrapped in `act` but now this is the norm.

Doing this investigation made me think about how we do `useSignalEffect` in React and maybe we should leverage the `scheduler` package rather than injecting an arbitrary animation frame as React would better work with its own scheduler.

### DIFF
--- a/.changeset/ninety-beans-compare.md
+++ b/.changeset/ninety-beans-compare.md
@@ -1,0 +1,5 @@
+---
+"@preact/signals": minor
+---
+
+Defer all DOM updates by an animation frame as well
--- a/.changeset/ninety-beans-compare.md
+++ b/.changeset/ninety-beans-compare.md
@@ -3,14 +3,14 @@
 ---
 
 Defer all DOM updates by an animation frame, this should make it so
-that any used to be synchronous DOM update will be delayed by an
+that any previously synchronous DOM update will be instead delayed by an
 animation frame. This allows Preact to first perform its own render
 cycle and then our direct DOM updates to occur. These will now
 be performed in a batched way which is more performant as the browser
 is prepared to handle these during the animation frame.
 
-This does have one way to how Preact based signals are tested, when
-we perform a signal update we have to wrap it in `act`, in a way
-this was always the case as a signal update that resulted in
-a Preact state update would require it to be wrapped in `act` but
+This does impact how Preact based signals are tested, when
+you perform a signal update, you'll need to wrap it in `act`. In a way
+this was always the case, as a signal update that resulted in
+a Preact state update would require it to be wrapped in `act`, but
 now this is the norm.
--- a/.changeset/ninety-beans-compare.md
+++ b/.changeset/ninety-beans-compare.md
@@ -1,5 +1,5 @@
 ---
-"@preact/signals": minor
+"@preact/signals": major
 ---
 
 Defer all DOM updates by an animation frame, this should make it so
--- a/.changeset/ninety-beans-compare.md
+++ b/.changeset/ninety-beans-compare.md
@@ -2,4 +2,15 @@
 "@preact/signals": minor
 ---
 
-Defer all DOM updates by an animation frame as well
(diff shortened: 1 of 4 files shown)


.changeset/ninety-beans-compare.md
```diff
@@ -0,0 +1,16 @@
+---
+"@preact/signals": minor
+---
+
+Defer all DOM updates by an animation frame, this should make it so
+that any used to be synchronous DOM update will be delayed by an
```
rschristian, line 6:
```suggestion
that any previously synchronous DOM update will be instead delayed by an
```

.changeset/ninety-beans-compare.md
```diff
@@ -0,0 +1,16 @@
...
+animation frame. This allows Preact to first perform its own render
+cycle and then our direct DOM updates to occur. These will now
+be performed in a batched way which is more performant as the browser
+is prepared to handle these during the animation frame.
+
+This does have one way to how Preact based signals are tested, when
```
rschristian, line 12: Sorry, not quite sure what's meant here but I think there's a typo. "This does impact how..." perhaps?

.changeset/ninety-beans-compare.md
```diff
@@ -0,0 +1,16 @@
...
+we perform a signal update we have to wrap it in `act`, in a way
+this was always the case as a signal update that resulted in
+a Preact state update would require it to be wrapped in `act` but
```
rschristian, line 15: The switch from "we" to "you" is mostly just to emphasize this is a change end users need to take. I think "we" isn't the best word for that as users might think that's an internal change, though I understand your meaning ("we" collectively as users of the lib)

```suggestion
you perform a signal update, you'll need to wrap it in `act`. In a way
this was always the case, as a signal update that resulted in
a Preact state update would require it to be wrapped in `act`, but
```

rschristian (approved):
Awesome!

.changeset/ninety-beans-compare.md
JoviDeCroock:
```suggestion
This does impact how Preact based signals are tested, when
```

marvinhagemeister (approved):
This is great! Only thought I have is if we should flag this update as "major" for the preact adapter or not. Because upgrading might break some tests.

.changeset/ninety-beans-compare.md
```diff
@@ -0,0 +1,16 @@
+---
+"@preact/signals": minor
```
JoviDeCroock, line 2:
```suggestion
"@preact/signals": major
```

PR #610, opened 2024-10-08
Throw error when binding signal to event handler

### LINKED ISSUE #607: Signals passed to event-handlers stop updating

**Environment**
- [x] I am using `@preact/signals`
- [ ] I am using `@preact/signals-react`
  - [ ] I understand usage changed in v2, and I've followed the [React Integration instructions](https://github.com/preactjs/signals/tree/main/packages/react#react-integration)

**Describe the bug**

When we pass a signal to an event-handler, the event-handler does not update when the signal updates.

**To Reproduce**

https://codesandbox.io/p/sandbox/quizzical-dubinsky-2t6fph?file=%2Fsrc%2Findex.js%3A3%2C10&workspaceId=2e3797ed-dfbc-47a1-a0f6-cc1df3dc9874

After 5 seconds the event-handler will switch to one where you type it will change the value to bonk.

**Expected behavior**

It should not invoke the old handler (see logs).

To fix this we might need to opt out event handlers from the optimized update path.

JoviDeCroock:
Resolves #607

### DIFF
--- a/.changeset/chilly-planets-learn.md
+++ b/.changeset/chilly-planets-learn.md
@@ -1,0 +1,5 @@
+---
+"@preact/signals": minor
+---
+
+Throw error when binding a signal to an event handler
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -193,12 +193,24 @@ hook(OptionsTypes.DIFFED, (old, vnode) => {
 				updaters = {};
 				dom._updaters = updaters;
 			}
+
 			for (let prop in props) {
 				let updater = updaters[prop];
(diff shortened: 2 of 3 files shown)


rschristian:
Sorry, had somehow missed this PR. Seems reasonable, though I wish we had a better detection method than starting with `'on'` (totally aware this is what Preact does too, matching it makes sense)

JoviDeCroock:
@rschristian no worries, yes my main reason for removing this is that the detection is flawed and we could in theory start using `handleEvent` as well to capture these 😅 though that would only be supported in 10.25.0

PR #611, opened 2024-10-08
fix(signals-react): use isomorphic layout

### LINKED ISSUE #606: Warning: useLayoutEffect does nothing on the server

**Environment**
- [x] I am using `@preact/signals-react`
  - [x] I understand usage changed in v2, and I've followed the [React Integration instructions](https://github.com/preactjs/signals/tree/main/packages/react#react-integration)

**Describe the bug**
Hello !
When running `@preact/signals-react` on a server-side rendered inertia app, I get the following warning:
```
Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. See https://reactjs.org/link/uselayouteffect-ssr for common fixes.
```
I guess it's rather an SSR problem than an Inertia problem.

It seems to be introduced by @developit in https://github.com/preactjs/signals/commit/0bed42d71929d26ecfdc1e48a16c92a0e6280563 commit. I'm not familiar enough with the package to really know what's the best way to avoid this warning.
For now, the only workaround I found is to go back to 2.0.1 version.

Let me now if you need more informations!

**To Reproduce**
No easy way I think.

**Expected behavior**
No warning.

Xstoudi:
Proposal to fix #606

### DIFF
--- a/.changeset/great-geese-play.md
+++ b/.changeset/great-geese-play.md
@@ -2,4 +2,4 @@
 "@preact/signals-react": patch
 ---
 
-Use isomorphic layout effect hook
+Silences noisy warnings about `useLayoutEffect` whilst using SSR by switching to an isomorphic layout effect hook
--- a/.changeset/great-geese-play.md
+++ b/.changeset/great-geese-play.md
@@ -1,0 +1,5 @@
+---
+"@preact/signals-react": patch
+---
+
+Use isomorphic layout effect hook
--- a/packages/react/runtime/src/index.ts
+++ b/packages/react/runtime/src/index.ts
@@ -316,6 +316,9 @@ function cleanupTrailingStore() {
 	currentStore?.f();
 }
 
+const useIsomorphicLayoutEffect =
+	typeof window !== "undefined" ? useLayoutEffect : useEffect;
+
 /**
  * Custom hook to create the effect to track signals used during render and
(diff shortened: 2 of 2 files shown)


rschristian (approved):
Cheers, mind adding a changeset? Should just need to run `pnpm changeset`, which will walk you through the process.

Xstoudi:
First time doing that, let me know if it's not alright.

.changeset/great-geese-play.md
```diff
@@ -0,0 +1,5 @@
+---
+"@preact/signals-react": patch
+---
+
+Use isomorphic layout effect hook
```
rschristian, line 5: Thanks!

We should add better docs here (I now realize this is completely missing from our [contributing doc](https://github.com/preactjs/signals/blob/main/CONTRIBUTING.md)), but the general idea is that this makes up our change log when we publish a new version.

As such, I'd suggest this as it makes it a bit clearer to users how it affects them:

```suggestion
Silences noisy warnings about `useLayoutEffect` whilst using SSR by switching to an isomorphic layout effect hook
```

rschristian:
Thanks! We appreciate you taking the time to both report & fix this.

PR #617, opened 2024-11-19
Fix text update recomputing in @preact/signals 1.3.x

developit:
We could have just checked the value against a local variable, but that would leave the computeds re-running unnecessarily.

Note that this would be patch - `@preact/signals@1.3.1`, as main is already at least a minor version ahead (or major).


JoviDeCroock (commented):
This needs a changeset, can look at CI failure later. Signals uses automated publishing but we could do a manual one for this

PR #627, opened 2024-12-13
Introduce react-utils package

JoviDeCroock:
This adds some common utilities for leveraging JSX with signals, I tried to stay in line with how `solid` provides JSX utils.


pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: '9.0'
```
JoviDeCroock, line 1: Latest lockfile version arrived, who dis

packages/react/utils/src/index.ts
```diff
@@ -0,0 +1,45 @@
...
+
+interface ForProps<T> {
+	each: Signal<Array<T>> | ReadonlySignal<Array<T>>;
+	fallback?: JSX.Element;
+	children: (value: T, index: number) => JSX.Element;
+}
+
+export function For<T>(props: ForProps<T>): JSX.Element | null {
+	useSignals();
+	const cache = useMemo(() => new Map(), []);
```
JoviDeCroock, line 29: We might need to switch this to a WeakMap when we see the first item being an object or leverage JSON.stringify on objects

packages/react/utils/src/index.ts
```diff
@@ -0,0 +1,45 @@
+import { ReadonlySignal, Signal } from "@preact/signals-core";
+import { useSignal } from "@preact/signals-react";
+import { useSignals } from "@preact/signals-react/runtime";
+import { Fragment, createElement, useMemo } from "react";
+
+interface ShowProps<T = boolean> {
+	when: Signal<T> | ReadonlySignal<T>;
+	fallback?: JSX.Element;
```
rschristian, line 8: I think we need to switch to importing `JSX`, it's (finally!) no longer a global in React 19 AFAIK. It should exist on the React namespace back to 16.

rschristian (commented):
I might've missed something but is there a particular reason we're offering these specifically and only for React?

JoviDeCroock:
@rschristian we will offer it for both but started out with React as the 2.0 release was pending

marvinhagemeister (approved):
LGTM, nice work! Let's get the CI passing and then it's ready to go in imo.

rschristian (approved):
Looks great to me too!

PR #628, opened 2024-12-15
fix: Replace body so fast-refresh function caching works

### LINKED ISSUE #626: Including @preact/signals-react-transform breaks rerender on hook change for fast refresh
- [ ] Check if updating to the latest version resolves the issue

**Environment**
- [X] I am using `@preact/signals-react`
  - [X] I understand usage changed in v2, and I've followed the [React Integration instructions](https://github.com/preactjs/signals/tree/main/packages/react#react-integration)

**Describe the bug**
React Fast Refresh has two modes of updating the view: 
1. updating the current component tree 
2. re-rendering the components from scratch

Most operations result in updating the current component tree but some operations, like adding or removing hooks, causes everything to be re-rendered and state to be lost. This is mandatory as trying to call a different number of hooks during rendering will cause React to fail with `Error: Rendered fewer hooks than expected. This may be caused by an accidental early return statement.` or `Error: Should have a queue. This is likely a bug in React. Please file an issue.` or something similar

When adding `@preact/signals-react-transform`, it seems like all operations result in trying to update to current component tree and thus modifying anything related to hooks shows one of the mentioned errors.

**To Reproduce**
https://github.com/Artur-/react-signals-hmr

1. `npm run dev`
5. Open localhost:5173
6. Click the counter a few times to change the state
7. Edit src/App.tsx and add a new `useState` hook

The `main` branch includes `@preact/signals-react-transform` and shows the problem
The `without-signals` branch shows how it is supposed to work

**Expected behavior**
The page should be updated and the counter reset to 0.
Instead, the page is empty and console contains the React error messages

Artur-:
Fixes #626

### DIFF
--- a/packages/react-transform/test/node/index.test.tsx
+++ b/packages/react-transform/test/node/index.test.tsx
@@ -669,6 +669,7 @@ describe("React Signals Babel Transform", () => {
 				};
 			`);
 
+			scope.path.scope.crawl();
 			const signalsBinding = scope.bindings["_useSignals"];
 			expect(signalsBinding).to.exist;
 			expect(signalsBinding.kind).to.equal("module");
--- a/.changeset/thirty-lies-destroy.md
+++ b/.changeset/thirty-lies-destroy.md
@@ -1,0 +1,8 @@
+---
+"@preact/signals-react-transform": patch
+---
+
+Avoid cloning the top-level component function when we are
+prepending `useSignals`. This fixes compatability with fast-refresh
+as it requires the function identity to correctly leverage its
+cache.
--- a/packages/react-transform/src/index.ts
+++ b/packages/react-transform/src/index.ts
@@ -342,11 +342,10 @@ function wrapInTryFinally(
 	path: NodePath<FunctionLike>,
 	state: PluginPass,
 	hookUsage: HookUsage
-): FunctionLike {
+): BabelTypes.BlockStatement {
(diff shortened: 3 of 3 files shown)


packages/react-transform/test/node/index.test.tsx
```diff
@@ -669,6 +669,7 @@ describe("React Signals Babel Transform", () => {
 				};
 			`);
 
+			scope.path.scope.crawl();
```
Artur-, line 672: According to https://github.com/babel/babel/issues/8358, the scope bindings are not automatically updated when modifying the ast - this seems to update them

JoviDeCroock (approved):

andrewiggins (approved):
Thanks!

Artur-:
Any chance of getting this into a release?

JoviDeCroock:
It's been merged for < 24 hours, we'll get to it when we do... I have to first test the signals preact 2.0 release that is attached to this and then publish. I'll try to get this done today but can't make promises

JoviDeCroock:
@Artur- nevermind, I forgot I [already tested the Preact stuff out](https://stackblitz.com/edit/vitejs-vite-hm7utb?file=src%2FTodos.tsx) but... just forgot.... 😅 the todo-list has been a bit chaotic, it's all released now, thank you again for the contribution!

Artur-:
Thank you very much, remember to calm down and take a break for the holidays :)

PR #630, opened 2024-12-20
Remove `HAS_HOOKS_STATE` bit

JoviDeCroock:
This is a thought exercise as a result of a fix in the Preact respository, strap in as this might become a doozy, open up a [Stackblitz](https://stackblitz.com/edit/vitejs-vite-kaaehu) and follow along. In 10.25.0 we fixed an issue where the hooks-state settling logic would prevent a component having both signals as well as state from re-rendering, [The relevant PR](https://github.com/preactjs/preact/pull/4560).

The issue was that when a component has a signal that updates and hooks that end up at the same value the component would _not_ rerender. 

```tsx
import { useEffect, useState } from 'preact/hooks';
import { useSignal } from '@preact/signals';

// In 10.25.x this will correctly rerender.
// In 10.12.x-10.24.x and before this won't rerender.
let rootRenders = 0;
export function App() {
  const [count, setCount] = useState(0);
  const signal = useSignal('hello world');

  useEffect(() => {
    setCount(1);
    setCount(0);
    signal.value = 'bye';
  }, []);

  return (
    <div>
      <p>Count: {count} - App Rendered {rootRenders++} times</p>
      <p>{signal.value}</p>
      <Signal />
    </div>
  );
}

let renders = 0;
function Signal() {
  const signal = useSignal('hello world');
  return (
    <p>
      {signal} - Signal Rendered {renders++} times
    </p>
  );
}
```

This was solved by calling the previous `shouldComponentUpdate` (in this case the global sCU set by `@preact/signals`) when the state settling logic decides to bail out. From a plugin perspective this is the right course of action as we want all plugins to have their lifecycles respected.

Now this does however bring us to the... interesting part, when a state hook now decides in 10.25.x that it needs to bail we'll rerender it anyway because we see `HAS_HOOK_STATE`.

Let's look at this in action

```tsx
import { useEffect, useState } from 'preact/hooks';
import { useSignal } from '@preact/signals';
import './app.css';

// In 10.25.x the App will rerender despite settling on the same value
// In 10.12.x-10.24.x and before this won't rerender as our state settles.
let rootRenders = 0;
export function App() {
  const [count, setCount] = useState(0);

  useEffect(() => {
    setCount(1);
    setCount(0);
  }, []);

  return (
    <div>
      <p>
        Count: {count} - App Rendered {rootRenders++} times
      </p>
      <p>{signal.value}</p>
      <Signal />
    </div>
  );
}

let renders = 0;
function Signal() {
  const signal = useSignal('hello world');
  return (
    <p>
      {signal} - Signal Rendered {renders++} times
    </p>
  );
}
```

Now what I want to achieve is a middle ground where we can respect hooks state settling as well as rerender correctly.

Another variable to consider with signals is [`hookState._component.props !== props;`](https://github.com/preactjs/preact/blob/42e7b450933134d66a871bb5370c702ef18cf3cc/hooks/src/index.js#L252), this line is meant to respect the top-down rerender flow. When the parent initiates a render and passes through an updated prop we want to rerender deeply as this is the classic state based Virtual DOM way. In signals we do shallow equality, this however means that any component having both signals and state-based VDOM will respect the top-down render which I guess is a "one up" for removing this `HAS_HOOKS_STATE`.

Now the question becomes how will this behave?

**Pre 10.12**

- State settles to the same value --> we rely on `HAS_HOOKS_STATE` to rerender the component
  - This would break :(
- Signal is updated in a hooks component --> we rely on `HAS_PENDING_UPDATE` to rerender the component
  - This would keep working :) 

**10.12-10.24**

- State settles to the same value --> signal sCU gets called and triggers a rerender with HAS_HOOKS_STATE
- Signal is updated in a hooks component
  - The underlying sCU is called if the state has no pending updates

**10.25.x and onwards** 

- State settles to the same value --> signal sCU would not get called
  - Can ignore any state update coming from signals when the state is changed but settles on the same value
- Signal is updated in a hooks component --> we rely on `HAS_PENDING_UPDATE` to rerender the component

### DIFF
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -30,7 +30,6 @@ export {
 };
 
 const HAS_PENDING_UPDATE = 1 << 0;
-const HAS_HOOK_STATE = 1 << 1;
 const HAS_COMPUTEDS = 1 << 2;
 
 // Install a Preact options hook
@@ -300,13 +299,6 @@ hook(OptionsTypes.UNMOUNT, (old, vnode: VNode) => {
 	old(vnode);
 });
 
-/** Mark components that use hook state so we can skip sCU optimization. */
-hook(OptionsTypes.HOOK, (old, component, index, type) => {
-	if (type < 3 || type === 9)
-		(component as AugmentedComponent)._updateFlags |= HAS_HOOK_STATE;
-	old(component, index, type);
-});
-
 /**
  * Auto-memoize components that use Signals/Computeds.
  * Note: Does _not_ optimize components that use hook/class state.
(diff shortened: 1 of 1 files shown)


JoviDeCroock:
Update, we should remove the base rerender when there are no signals to also account for hooks-state as seen in https://github.com/preactjs/signals/pull/630/commits/2c97a21974fc7915509d6b91b0f1dc68ec78ff75 else state settling in hooks would never be respected.

packages/preact/src/index.ts
```diff
@@ -320,38 +320,28 @@ Component.prototype.shouldComponentUpdate = function (
...
-
-	// if there is a pending re-render triggered from Signals,
-	// or if there is hook or class state, update:
-	if (this._updateFlags & (HAS_PENDING_UPDATE | HAS_HOOK_STATE)) return true;
-
+	// If this is a component using state, rerender
 	// @ts-ignore
 	for (let i in state) return true;
 
+	if (this.__f || (typeof this.u == "boolean" && this.u === true)) {
```
JoviDeCroock, line 327: I'd love to have opinions on this retroactive fix, with adding `this.u` here we don't necessarily do anything wrong, when hooks would settle into a falsey update they would never call this function before so as of 10.25.x it fixes an issue and in pre 10.24.x it doesn't really do anything

marvinhagemeister (approved):

Issue #428, opened 2023-10-31
Adding a callback that will be triggered when the signal starts to be observed


szagi3891:
This callback could behave like this:

```
const exchangeRate = signal(0,

    // Callback triggered when the signal begins to observe something.
    () => {

        const timer = setTimeout(async () => {

            // Fetch the current exchange rate (of type number) using a request (fetch).
            const currentExchangeRate: number = await fetch( .... );

            // Set the exchange rate signal's value to the current rate.
            exchangeRate.value = currentExchangeRate
        });

        // Callback triggered when the signal stops observing something.
        return () => {

            // Clear the timer.
            clearInterval(timer);
        };
    }
);
```

For example, MobX implements it like this:
https://mobx.js.org/lazy-observables.html

mbeckem:
The [TC39 Signals proposal](https://github.com/tc39/proposal-signals) includes a similar feature:

```ts
    interface SignalOptions<T> {
        // Custom comparison function between old and new value. Default: Object.is.
        // The signal is passed in as the this value for context.
        equals?: (this: Signal<T>, t: T, t2: T) => boolean;

        // Callback called when isWatched becomes true, if it was previously false
        [Signal.subtle.watched]?: (this: Signal<T>) => void;

        // Callback called whenever isWatched becomes false, if it was previously true
        [Signal.subtle.unwatched]?: (this: Signal<T>) => void;
    }
```

`watched` and `unwatched` can be used, for example, to start (or stop) observing a foreign data source on demand, only when the signal is actually being used.

Issue #351, opened 2023-04-19
Subscription integration for third-party libraries


nahudalla:
Hello, I'm experimenting with Preact and Signals on a new side project I started recently. I really like the simplicity and speed so far.

**TL;DR:** I'd like to ask if it's possible to change something, perhaps add a new "hooks" functionality to "hook into" the subscription mechanism of the Signal class.

For this new side project, I'm experimenting with new ways to separate as much as possible my app's logic and data fetching from the UI layer. I think Signals are a great way of accomplishing jus that, however I'd also like non-UI code to remain framework- and library-agnostic.

Because of that, I'm trying to use Preact's Signals in my UI layer but I'm experimenting with the standard [EventTarget API](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget) for my non-UI code. This brings me to the point of this issue, which is that there's currently no way to detect when a new subscriber is added or removed from a Signal. I need this functionality to be able to add and remove subscriptions to external event sources only when the Signal is actually being used, and thus avoiding dangling subscriptions.

Currently, I'm trying to expand the default Signal functionality by extending the class. By looking at [this package's source code](https://github.com/preactjs/signals/blob/main/packages/core/src/index.ts#L212-L216) I've noticed that there are two internal methods that I'd like to override (`_subscribe` and `_unsubscribe`) to detect when a subscription is added and removed. However, since these methods are marked as `@internal`, the minified version of the library has different (shorter) names for these specific methods and therefore cannot be overridden by using their original name.

I'd like to ask if it's possible to change something, perhaps add a new "hooks" functionality to "hook into" the subscription mechanism of the Signal class. I'd be more than happy to help with the implementation of this new feature and create a PR but since I haven't worked with this code before, maybe I'm missing something and what I want shoulnd't be done for some reason.

PD: Sorry for the wall of text!

rschristian:
> However, since these methods are marked as @internal, the minified version of the library has different (shorter) names for these specific methods and therefore cannot be overridden by using their original name.

We use a `mangle.json` file to ensure consistent mangling. [They're always converted to `S` and `U` respectively](https://github.com/preactjs/signals/blob/89b109ab89cbf0cff379eb37f17371319e5034c6/mangle.json#L61-L62f). Patch them as you please.

nahudalla:
[...]

PR #634, opened 2024-12-27
Add watched and unwatched

JoviDeCroock:
Resolves https://github.com/preactjs/signals/issues/351
Resolves https://github.com/preactjs/signals/issues/428

This adds a second argument to `signal` which are the `options`, currently this adds two properties, watched and unwatched. These will most likely serve third party libraries a lot more than first party consumers.

- watched will trigger when the first subscriber subscribes to a signal
- unwatched will trigger when the last subscriber unsubscribes from a signal

The second parameter can in the future be used for other RFC's in our pipeline like the equality function. The addition of this parameter is in line with the TC39 proposal for browser-native signals https://github.com/tc39/proposal-signals

I opted not to add this to `computed` for the time being as the signal itself is the atom while computed is more of a derived value. Not sure if a utility like this would be useful there. In the TC39 proposal they also add this to computed, so might as well for parity.

Note that calling the `watcher` on the `subscribe` is slightly distinct from how we normally treat signals, i.e. we calculate the `computed` when its value is accessed for the first time.

### DIFF
--- a/.changeset/tough-rice-cover.md
+++ b/.changeset/tough-rice-cover.md
@@ -1,0 +1,7 @@
+---
+"@preact/signals-core": minor
+"@preact/signals": minor
+"@preact/signals-react": minor
+---
+
+Add an option to specify a watched/unwatched callback to a signal
--- a/README.md
+++ b/README.md
@@ -20,21 +20,16 @@ npm install @preact/signals-react
 npm install @preact/signals-core
 ```
 
-- [Guide / API](#guide--api)
-  - [`signal(initialValue)`](#signalinitialvalue)
-    - [`signal.peek()`](#signalpeek)
-  - [`computed(fn)`](#computedfn)
-  - [`effect(fn)`](#effectfn)
-  - [`batch(fn)`](#batchfn)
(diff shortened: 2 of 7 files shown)


andrewiggins:
Is this mostly to align with the signal proposal? If so, perhaps we can wait for that proposal to go to Stage 2 or 3 and we have more confidence in the API landing before we implement (just to avoid any churn).

if we have some use cases we'd like to use this for now, it does seems reasonable to me.

mxck:
> Is this mostly to align with the signal proposal? If so, perhaps we can wait for that proposal to go to Stage 2 or 3 and we have more confidence in the API landing before we implement (just to avoid any churn).
> 
> if we have some use cases we'd like to use this for now, it does seems reasonable to me.

Hi! I’m one of the people who requested this feature.

The main use case is for subscriptions over WebSocket or data polling using timeout, but only when the signal is mounted/used.

I previously asked about how we could achieve a setup where data updates occur only if the signal is actively in use. Here’s the link to my question in discussions: [link](https://github.com/preactjs/signals/discussions/635).

marvinhagemeister (approved):
LGTM

PR #640, opened 2025-01-07
[React] Don't track signals read during SSR render

andrewiggins:
Currently during all renders we setup an effect to track which signals a component reads, so we can rerender that component when a signal it uses changes. When that component unmounts, React calls our unsubscribe function and we dispose of the created effects.

However, during an SSR render, React does not call our unsubscribe function. Internally we use `useSyncExternalStore` to hook up React to signals. Subscribing/unsubscribing to stores is treated as a passive effect (aka like a `useEffect`), which aren't invoked on SSR renders. So while we setup our effect to track signal usage during SSR, the corresponding unsubscribe method is never called at the end of SSR so the tracking objects are never cleaned up from the Signal.

In practice, each tracking object is ~100 bytes and so many applications may not notice this if they deploy frequently. You may notice GC times taking longer first before memory pressure. Regardless, this PR fixes it.

This PR fixes the memory leak by not creating any tracking effect if the `window` object is `undefined`.


packages/react/runtime/src/index.ts
```diff
@@ -312,7 +339,11 @@ export function _useSignalsImplementation(
 
 	const storeRef = useRef<EffectStore>();
 	if (storeRef.current == null) {
-		storeRef.current = createEffectStore(_usage);
+		if (typeof window === "undefined") {
```
JoviDeCroock, line 342: The thing I am worried about here is react-native and others, shouldn't we detect this with i.e. layout-effect/effect firing?

JoviDeCroock: I guess it's not really possible so I'd add a global so we have an escape hatch, like a function `enableX` which defaults to the typeof window check

JoviDeCroock (approved):
Leaving my approval to unblock, pending the lint fix and maybe the global override for some environments

packages/react/runtime/src/index.ts
andrewiggins: Hmmm true true...

andrewiggins: Yea that may be a decent escape hatch, but Imma sit on this for a day or two more to see if some other solution presents itself.

What I'd really like is some hook or something that React would call as an effect in both SSR and client-side renders but I couldn't find one...

andrewiggins: It looks like React Native does define a global `window` object:
- https://stackoverflow.com/a/49911697
- https://github.com/facebook/react-native/blob/ec72af403ca6cc64a6c7e99b00a6a6b5d56ff2db/packages/react-native/Libraries/Core/setUpGlobals.js

So maybe this check will suffice?

brianlyn: Just popping in here to say I ran into an issue using preact signals with [react ink](https://github.com/vadimdemedes/ink) because of the lack of window object. Any chance we could get that escape hatch baked in?

PR #645, opened 2025-01-11
Upgrade pnpm

JoviDeCroock:
Every time I opened a branch I had a lot of pending changes, this intends to resolve that by bumping the pnpm version to 9.12

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Install pnpm
         uses: pnpm/action-setup@v3
         with:
-          version: 8
+          version: 9.23
 
       - name: Install Node.js
         uses: actions/setup-node@v4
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       - name: Install pnpm
         uses: pnpm/action-setup@v3
         with:
-          version: 8
+          version: 9.23
```
rschristian, line 25: Is this version meant to not matching compressed-size & main? Is there a reason we're not just using `9`?

rschristian (approved):
Thanks for doing this! I think we've all been ignoring the lockfile for a while 🙈

.github/workflows/release.yml
JoviDeCroock, line 25:
```suggestion
          version: 9.12
```

PR #660, opened 2025-03-26
Use afterNextFrame for hidden origins

JoviDeCroock:
When a page is hidden because the user is on a different tab but it received i.e. a webhook we still want effects to run. This PR ensures that by installing a 100ms timeout as animation frames don't run for non-visible pages.

### DIFF
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -388,10 +388,22 @@ export function useComputed<T>(compute: () => T) {
 	return useMemo(() => computed<T>(() => $compute.current()), []);
 }
 
+const HAS_RAF = typeof requestAnimationFrame == "function";
+function afterNextFrame(callback: () => void) {
+	let raf: number | undefined;
+
+	const done = () => {
+		clearTimeout(timeout);
+		if (HAS_RAF && raf != null) cancelAnimationFrame(raf);
+		setTimeout(callback);
+	};
+
+	const timeout = setTimeout(done, 100);
+	if (HAS_RAF) raf = requestAnimationFrame(done);
+}
+
 const deferEffects =
-	typeof requestAnimationFrame === "undefined"
-		? setTimeout
-		: requestAnimationFrame;
+	typeof requestAnimationFrame === "undefined" ? setTimeout : afterNextFrame;
 
 const deferDomUpdates = (cb: any) => {
 	queueMicrotask(() => {
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -394,8 +394,8 @@ function afterNextFrame(callback: () => void) {
 
 	const done = () => {
 		clearTimeout(timeout);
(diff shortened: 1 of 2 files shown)


packages/preact/src/index.ts
```diff
@@ -388,10 +388,22 @@ export function useComputed<T>(compute: () => T) {
...
+
+	const done = () => {
+		clearTimeout(timeout);
+		if (HAS_RAF && raf != null) cancelAnimationFrame(raf);
+		setTimeout(callback);
+	};
+
+	const timeout = setTimeout(done, 100);
+	if (HAS_RAF) raf = requestAnimationFrame(done);
+}
```
developit, line 403: nit: in the case where we fall back to the timeout, we don't need to run the nested `setTimeout(0)` to wait until after the frame - there's no frame.

other nit: `requestAnimationFrame()` never returns `0`, so we can just use `if (raf)` to guard `cancelAnimationFrame` (it represents `HAS_RAF && raf != null`).

```suggestion
const HAS_RAF = typeof requestAnimationFrame == "function";
function afterNextFrame(callback: () => void) {
	let raf: number | undefined;

	const done = () => {
		clearTimeout(timeout);
		if (raf) cancelAnimationFrame(raf);
		setTimeout(callback);
	};

	const timeout = setTimeout(callback, 100);
	if (HAS_RAF) raf = requestAnimationFrame(done);
}
```

JoviDeCroock: We still need to clear the `animation` frame if there is one hence why I am calling `done` rather than `callback` but yes, `done` should not invoke another callback

```suggestion
const HAS_RAF = typeof requestAnimationFrame == "function";
function afterNextFrame(callback: () => void) {
	let raf: number | undefined;

	const done = () => {
		clearTimeout(timeout);
		if (raf != null) cancelAnimationFrame(raf);
		callback();
	};

	const timeout = setTimeout(done, 100);
	if (HAS_RAF) raf = requestAnimationFrame(done);
}
```

packages/preact/src/index.ts
```diff
@@ -388,10 +388,22 @@ export function useComputed<T>(compute: () => T) {
...
+
+	const done = () => {
+		clearTimeout(timeout);
+		if (raf != null) cancelAnimationFrame(raf);
+		callback();
+	};
+
+	const timeout = setTimeout(done, 100);
+	if (HAS_RAF) raf = requestAnimationFrame(done);
+}
```
developit, line 403: I just realized we don't need `HAS_RAF` at all here, since this function is only used when `typeof requestAnimationFrame !== 'undefined'` due to line 405.

```suggestion
function afterNextFrame(callback: () => void) {
	const done = () => {
		clearTimeout(timeout);
		cancelAnimationFrame(raf);
		callback();
	};

	const timeout = setTimeout(done, 100);
	const raf = requestAnimationFrame(done)
}
```

developit (approved):

PR #674, opened 2025-05-03
Remove duplicate node version check and add type as Node | undefined for head

nishan-singh:
I removed duplicated node source version check in `needToRecompute` func and make the head's type `Node | undefined` in `cleanUpSources` func.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -412,8 +412,7 @@ function needsToRecompute(target: Computed | Effect): boolean {
 		// dependency cycle), then we need to recompute.
 		if (
 			node._source._version !== node._version ||
-			!node._source._refresh() ||
-			node._source._version !== node._version
+			!node._source._refresh()
 		) {
 			return true;
 		}
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -456,7 +456,7 @@ function prepareSources(target: Computed | Effect) {
 
 function cleanupSources(target: Computed | Effect) {
 	let node = target._sources;
-	let head = undefined;
+	let head: Node | undefined = undefined;
 
 	/**
 	 * At this point 'target._sources' points to the tail of the doubly-linked list.


packages/core/src/index.ts
```diff
@@ -457,7 +456,7 @@ function prepareSources(target: Computed | Effect) {
 
 function cleanupSources(target: Computed | Effect) {
 	let node = target._sources;
-	let head = undefined;
+	let head: Node | undefined = undefined;
```
JoviDeCroock, line 459:
```suggestion
	let head: Node;
```

This should be sufficient for TS

JoviDeCroock:
Your PR needs a changeset

packages/core/src/index.ts
```diff
@@ -457,7 +456,7 @@ function prepareSources(target: Computed | Effect) {
 
 function cleanupSources(target: Computed | Effect) {
 	let node = target._sources;
-	let head = undefined;
+	let head: Node;
```
JoviDeCroock, line 459: Apparently it's not, apologies 😅

nishan-singh:
I added a changeset. I hope I did it right. I'm really sorry if I messed it up. I have never contributed to open source before.

.changeset/chubby-parks-fetch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@preact/signals-core": minor
```
rschristian, line 2:
```suggestion
"@preact/signals-core": patch
```

rschristian (dismissed):

JoviDeCroock (approved):

rschristian:
Sorry for the delay here, been quite busy.

It looks like this change isn't quite as innocent as it appears, judging by the state of the CI. `._refresh()` will need to be called prior to the check of `._version` if the version matches, which this PR breaks.

We might be able to get away with removing the first check instead, though that might hurt benchmarks a bit. Would need to take a deeper look into that.

marvinhagemeister (changes requested):
This check is not redundant. It's a performance optimization. The first check is whether we need to recompute it at all. If the versions are already different then we don't need to walk the signal graph to check if it's dirty. We already know that.

But if there the same and we refresh the state of signals, there might still be no changes being necessary in the end. That is the case when the versions stay the same. That's why the same condition is checked again after that. It allows us to skip a bit of unnecessary work and is hence a performance optimization.

packages/core/src/index.ts
```diff
@@ -410,11 +410,7 @@ function needsToRecompute(target: Computed | Effect): boolean {
 		// If there's a new version of the dependency before or after refreshing,
 		// or the dependency has something blocking it from refreshing at all (e.g. a
 		// dependency cycle), then we need to recompute.
-		if (
-			node._source._version !== node._version ||
-			!node._source._refresh() ||
-			node._source._version !== node._version
-		) {
+		if (!node._source._refresh() || node._source._version !== node._version) {
```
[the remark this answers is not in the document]
nishan-singh: I removed the changeset and version check

rschristian (approved):

PR #679, opened 2025-05-21
Fix accessing signals from Preact Class Component constructor.

WeirdConstructor:
This pull request fixes a bug we discussed on slack today.

I was using signals with Preact in class components when I stumbled upon this error in the Browser: `Uncaught (in promise) TypeError: e.__ is null`. Where `__` is the mangled `_parent` property of a VNode. The TypeError occurs in the renderComponent() function of Preact, where it tries to access the parent to insert the newVNode.

The issue was, that reading the signal in the constructor of a top level component caused some internal Fragment VNode to be diffed. This change skips over the internal Fragment in the RENDER hook.

I needed to bump the devDependencies of preact from 10.9.0 to 10.25.0, because this error does only occur with a newer version of preact.

### DIFF
--- a/packages/preact/package.json
+++ b/packages/preact/package.json
@@ -50,7 +50,7 @@
     "preact": ">= 10.25.0"
   },
   "devDependencies": {
-    "preact": "10.25.0",
+    "preact": "^10.25.0",
     "preact-render-to-string": "^5.2.5"
   },
   "publishConfig": {
--- a/packages/preact/package.json
+++ b/packages/preact/package.json
@@ -50,7 +50,7 @@
     "preact": ">= 10.25.0"
   },
   "devDependencies": {
-    "preact": "10.9.0",
+    "preact": "10.25.0",
     "preact-render-to-string": "^5.2.5"
   },
   "publishConfig": {
--- a/packages/preact/package.json
+++ b/packages/preact/package.json
@@ -50,7 +50,7 @@
     "preact": ">= 10.25.0"
   },
   "devDependencies": {
-    "preact": "^10.25.0",
+    "preact": "^10.26.6",
     "preact-render-to-string": "^5.2.5"
   },
   "publishConfig": {
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -1,4 +1,4 @@
-import { options, Component, isValidElement } from "preact";
+import { options, Component, isValidElement, Fragment } from "preact";
 import { useRef, useMemo, useEffect } from "preact/hooks";
 import {
 	signal,
@@ -190,6 +190,12 @@ hook(OptionsTypes.DIFF, (old, vnode) => {
 
 /** Set up Updater before rendering a component */
 hook(OptionsTypes.RENDER, (old, vnode) => {
+	if (vnode.type === Fragment) {
+		// Ignore the Fragment inserted by preact.createElement().
+		old(vnode);
+		return;
+	}
+
 	setCurrentUpdater();
 
 	let updater;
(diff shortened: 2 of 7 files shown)


packages/preact/package.json
```diff
@@ -50,7 +50,7 @@
     "preact": ">= 10.25.0"
   },
   "devDependencies": {
-    "preact": "10.9.0",
+    "preact": "10.25.0",
```
JoviDeCroock, line 53: You forgot to update the lockfile

packages/preact/src/index.ts
```diff
@@ -190,6 +190,12 @@ hook(OptionsTypes.DIFF, (old, vnode) => {
 
 /** Set up Updater before rendering a component */
 hook(OptionsTypes.RENDER, (old, vnode) => {
+	if (vnode.type === Fragment) {
```
JoviDeCroock, line 193: Let's wrap the thing underneath in a `!== Fragment` to save bytes

JoviDeCroock:
This looks dangerous, please don't update all deps

WeirdConstructor:
> This looks dangerous, please don't update all deps

I have not yet worked a with pnpm. How do I update only the version of preact in the lockfile? Manual editing?

JoviDeCroock:
Just update the preact version to whatever you like for instance `^10.25.0` and run `pnpm i` at the root.

WeirdConstructor:
Ah, ok, will do that tomorrow.

WeirdConstructor:
The deploy-preview passes now, yay :-)

JoviDeCroock:
Awesome, can you just add a patch changeset with `pnpm changeset` or by clicking the prompt from the bot in the UI here

WeirdConstructor:
I hope the changeset is correct? And I tried looking into the issues eslint found. But I have no clue what to make of the errors it produces. Also, running `pnpm run lint:eslint` does not do anything locally here, producing no output at all, neither on my Linux or my Windows machine.

JoviDeCroock:
The errors are from the typecheck and are caused by the Preact upgrade, I can look at them CC @rschristian

rschristian (approved):

packages/preact/test/index.test.tsx
```diff
@@ -962,4 +968,37 @@ describe("@preact/signals", () => {
 			"<div><div>4</div><div>5</div><div>6</div></div>"
 		);
 	});
+
+	describe("Preact class Component", () => {
+		it("should support reading signal in class component constructor", async () => {
+			const count = signal(1);
+			const spy = sinon.spy();
+
+			class App extends Component {
```
JoviDeCroock, line 977:
```suggestion
			class App extends Component<{ x: number }, {}> {
```

packages/preact/test/index.test.tsx
```diff
@@ -962,4 +968,37 @@ describe("@preact/signals", () => {
...
+					spy("willmount:" + count.value);
+				}
+
+				render() {
+					return <div>{count}</div>;
+				}
+			}
+			expect(() =>
+				act(() => {
+					render(createElement(App, { x: 1 }), scratch);
```
JoviDeCroock, line 993:
```suggestion
					render(<App x={1} />, scratch);
```

JoviDeCroock (approved):

JoviDeCroock:
Thank you for starting this effort and unveiling this bug, great work!

JoviDeCroock:
@WeirdConstructor I made a lint error myself, do you have the bandwidth to solve this? If you want to verify locally https://github.com/preactjs/signals/blob/main/.github/workflows/main.yml#L36 the command can be found here

WeirdConstructor:
@JoviDeCroock yes, I am pulling the branch now and will try to fix that lint error.

PR #681, opened 2025-05-25
Debug signals

JoviDeCroock:
This creates an initial draft for having console based debug tooling for signals, you can name your signals and start debugging them by importing this package.

This can set us up for more visual tools or just generally a debugging story in signals. From here we can add two things which I would think of as valuable

- A transpiler plugin that auto-fills in names
- A good way for the Preact plugin to fill in i.e. `SignalValue` for the optimized updates

Example:

Screenshot 2025-05-25 at 16 31 34


packages/debug/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "src",
+    "dist",
+    "CHANGELOG.md",
+    "LICENSE",
+    "README.md"
+  ],
+  "scripts": {
+    "prepublishOnly": "cd ../.. && pnpm build:debug"
+  },
+  "dependencies": {
```
JoviDeCroock, line 43: These should be devDependencies

packages/debug/README.md
JoviDeCroock, line 1: still have to write a good readme

packages/debug/src/index.ts
```diff
@@ -0,0 +1,255 @@
...
+import { UpdateInfo, Node, Computed as ComputedType } from "./internal";
+
+const inflightUpdates = new Set<Signal>();
+const updateInfoMap = new WeakMap<Signal, UpdateInfo[]>();
+const trackers = new WeakMap<Signal, number>();
+const activeSignals = new Set<Signal>();
+const signalValues = new WeakMap<Signal, any>();
+const subscriptions = new WeakMap<Signal, () => void>();
+
+export function getDebugStats() {
```
JoviDeCroock, line 13: Probably worth removing this now, mainly used this to see whether we leaked memory

docs/demos/index.tsx
```diff
@@ -101,6 +101,7 @@ function DuelingCounters() {
 	);
 }
 
+window.location.href = "/demos";
```
developit, line 104:
```suggestion
```

packages/core/src/index.ts
```diff
@@ -343,17 +346,19 @@ Signal.prototype._unsubscribe = function (node) {
...
+			const value = this.value;
+			const prevContext = evalContext;
+			evalContext = undefined;
+			try {
+				fn(value);
+			} finally {
+				evalContext = prevContext;
+			}
+		},
+		{ name: this.name + "-subscribe" }
```
developit, line 360: nit: could this just be `"subscribe"` and the debug lib calculates the name? the effect will only ever have a single dependency.

PR #686, opened 2025-05-27
Add async computed with async callbacks

JoviDeCroock:
## Summary

- add an experimental core `asyncComputed()` primitive for ordinary synchronous or `async` callbacks
- expose reactive `pending`, `settled`, `failed`, and `error` state while retaining the last successful value and ignoring stale results
- make async computeds created by `createModel()` automatically follow model disposal
- add `useAsyncComputed()` to the Preact and React adapters, including error-boundary behavior and Suspense for externally/model-owned instances
- defer hook-owned callback work until commit so abandoned and server renders do not start it

## Design choice

This is the ordinary `async`/`await` option. Signal reads are tracked synchronously, so reactive inputs must be captured before the first `await`:

```ts
const user = asyncComputed(async () => {
  const id = userId.value;
  const response = await fetch(`/api/users/${id}`);
  return response.json();
});
```

The alternative [generator-based prototype](https://github.com/preactjs/signals/compare/main...JoviDeCroock%2Fsolid-async-learnings?body=&expand=1) re-enters dependency tracking after each yield. Keeping these separate lets us compare familiar async ergonomics against complete cross-suspension tracking.

Hook-created instances render their explicit state and cannot suspend on initial mount because framework hook state does not survive that suspension. For Suspense, create the async computed in a model or above the boundary and pass the instance to `useAsyncComputed(instance, { suspend: true })`.

### DIFF
--- a/.changeset/tough-rice-cover.md
+++ b/.changeset/tough-rice-cover.md
@@ -1,0 +1,7 @@
+---
+"@preact/signals-core": minor
+"@preact/signals": minor
+"@preact/signals-react": minor
+---
+
+Add an option to specify a watched/unwatched callback to a signal
--- a/.changeset/twelve-meals-tease.md
+++ b/.changeset/twelve-meals-tease.md
@@ -1,0 +1,5 @@
+---
+"@preact/signals": minor
+---
+
+Add `asyncComputed` primitive to have a first class util to handle (suspenseful) data fetching.
--- a/README.md
+++ b/README.md
@@ -20,21 +20,16 @@ npm install @preact/signals-react
 npm install @preact/signals-core
 ```
 
-- [Guide / API](#guide--api)
-  - [`signal(initialValue)`](#signalinitialvalue)
-    - [`signal.peek()`](#signalpeek)
-  - [`computed(fn)`](#computedfn)
-  - [`effect(fn)`](#effectfn)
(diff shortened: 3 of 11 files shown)


JoviDeCroock:
Yes, makes sense to explicitly document this. I thought about it but apparently did not write it 😂

JoviDeCroock:
The difference between this primitive and react-query is that it can be used for more than just `fetch`, consider webworker traffic and other use cases.

mxck:
True, it’s broader than `fetch`, but `AbortSignal` is also supported in streams, locks, and more. It’s especially useful for 2–3 chained `fetch` calls or other long tasks where you need to handle cancellation properly.

rschristian (approved):
Hype hype hype, probably one of the most requested features? Would be awesome to see!

> the issue being that this is based on `useId()` which isn't readily available in React 18 according to the types

[`useId` came out w/ React 18.0](https://react.dev/blog/2022/03/29/react-v18#useid), types are probably incorrect there.

> Question: When the input variables change for `computed` would you expect us to erase `data` and `error` or should we retain it and fetch in the background? Alternatively we can add a `fetching` signal to have a background indicator.
>
> EDIT: Added `running` as a way to differentiate between things

That sounds like a great solution to me.

Issue #671, opened 2025-05-01
Redundant condition in version check logic


nishan-singh:
Hi,

I was reading the source code of preact signals (https://github.com/preactjs/signals/blob/main/packages/core/src/index.ts) and I noticed a small redundancy in the code. 

![Image](https://github.com/user-attachments/assets/56131ce2-d2d8-4db5-b45e-1f5dbff58e65)

The condition `node._source._version !== node._version` appears twice in the if statement. I am not an expert, but I think, the second condition could be removed. 

One more thing:

Typescript complains that head is of type `undefined` at line 498 (cleanUpSources func). I will just paste the error here:

`index.ts|498 col 4-8 error| Type 'Node' is not assignable to type 'undefined'.`

I think, an easy fix would be to make head have `Node | undefined` type.

![Image](https://github.com/user-attachments/assets/e267aae1-dfc4-4b82-a998-c4ecdb05fa33)

I would be happy to make a pull request  to fix these minor issues and contribute to open source for the first time (well, not really, I tried to contribute to a project but it didn't go well lol)

rschristian:
> The condition node._source._version !== node._version appears twice in the if statement. I am not an expert, but I think, the second condition could be removed.

Nice spot! Likely a copy/paste or merge artifact left around.

PRs are very welcome and appreciated if you'd like to correct one or both issues. Thanks!

nishan-singh:
I just submitted a pull request!

nishan-singh:
Hi @rschristian,

Sorry to bother you, but I wanted to ask if there's anything else I need to do as the PR is still open?

JoviDeCroock:
You need to add a changeset

JoviDeCroock:
As pointed out by Jviide in the [PR](https://github.com/preactjs/signals/pull/674#issuecomment-2923154488) this check is intentional. I should have seen the comment above it where it explicitly sais we want to check before and after, my sincerest apologies @nishan-singh

nishan-singh:
I should have seen that too. I apologize for opening up this issue and PR without reading the comment 🤦.

PR #690, opened 2025-05-30
Modify comments explaining why dependency versions are checked before and after refresh

jviide:
There have been some questions (e.g. #671) why the internal `needsToRecompute` function in the core checks dependency versions before _and_ after refreshing the dependency. This pull request tries to address this by modifying the surrounding code comments to make the reasoning a bit more explicit.

The first check is not strictly necessary for the functionality, because an incremented version number will stay incremented, but it allows skipping the `_refresh()` call in cases where the dependency has already definitely been updated.

Big thanks to @nishan-singh for pointing out the ambiguousness of the code.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -433,12 +433,16 @@ function needsToRecompute(target: Computed | Effect): boolean {
 		node !== undefined;
 		node = node._nextSource
 	) {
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
LGTM

PR #691, opened 2025-06-03
allow jsx alternatives

robin-drexler:
When we are dealing with transpiled JSX, be that because of a library or a component that is just written without JSX, the babel plugin will fail to notice these components and not inject `useSignals`. The absence of `useSignals` makes these components non-reactive to updates to the signals within the component.

This adds a new option to check for `jsx`/`createElement`, when components have them we treat it the same as we'd find JSX and register the component for reactivity.

### DIFF
--- a/packages/react-transform/src/index.ts
+++ b/packages/react-transform/src/index.ts
@@ -24,6 +24,8 @@ const getHookIdentifier = "getHookIdentifier";
 const maybeUsesSignal = "maybeUsesSignal";
 const containsJSX = "containsJSX";
 const alreadyTransformed = "alreadyTransformed";
+const jsxIdentifiers = "jsxIdentifiers";
+const jsxObjects = "jsxObjects";
 
 const UNMANAGED = "0";
 const MANAGED_COMPONENT = "1";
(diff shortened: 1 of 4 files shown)


developit (approved):

packages/react-transform/src/index.ts
```diff
@@ -577,6 +688,14 @@ export default function signalsTransform(
 			FunctionDeclaration: visitFunction,
 			ObjectMethod: visitFunction,
 
+			CallExpression(path, state) {
+				if (options.experimental?.detectJSXAlternatives) {
```
JoviDeCroock, line 692: Let's just call this options.detectTransformedJsx

packages/react-transform/src/index.ts
```diff
@@ -330,6 +332,33 @@ function isValueMemberExpression(
...
+	if (
+		path.node.callee.type === "MemberExpression" &&
+		path.node.callee.object.type === "Identifier" &&
+		path.node.callee.property.type === "Identifier"
+	) {
+		const objectName = path.node.callee.object.name;
+		const methodName = path.node.callee.property.name;
+		const allowedMethods = jsxObjectMap?.get(objectName);
+		return allowedMethods?.includes(methodName) ?? false;
+	}
```
developit, line 357:
```suggestion
  const callee = path.get("callee");

	// Check direct function calls like _jsx("div", props) or createElement("div", props)
	if (callee.isIdentifier()) {
		return jsxIdentifierSet?.has(callee.node.name) ?? false;
	}

	// Check member expression calls like React.createElement("div", props) or jsxRuntime.jsx("div", props)
	if (
		callee.isMemberExpression() &&
		callee.get("object").isIdentifier() &&
		callee.get("property").isIdentifier()
	) {
		const objectName = callee.node.object.name;
		const methodName = callee.node.property.name;
		const allowedMethods = jsxObjectMap?.get(objectName);
		return allowedMethods?.includes(methodName) ?? false;
	}
```

packages/react-transform/src/index.ts
```diff
@@ -465,6 +494,78 @@ function createImportLazily(
...
+			}
+		},
+		VariableDeclarator(varPath) {
+			if (
+				varPath.node.init?.type === "CallExpression" &&
+				varPath.node.init.callee.type === "Identifier" &&
+				varPath.node.init.callee.name === "require" &&
+				varPath.node.init.arguments[0].type === "StringLiteral"
+			) {
+				const packageName = varPath.node.init.arguments[0].value;
```
developit, line 540:
```suggestion
      const init = varPath.get("init");
			if (
				init.isCallExpression() &&
				init.get("callee").isIdentifier() &&
				init.get("callee").node.name === "require" &&
				init.get("arguments.0").isStringLiteral()
			) {
				const packageName = init.node.arguments[0].value;
```

developit (approved):

packages/react-transform/test/node/index.test.tsx
```diff
@@ -676,4 +676,265 @@ describe("React Signals Babel Transform", () => {
...
+				expectedOutput,
+				{
+					experimental: { detectJSXAlternatives: true },
+				},
+				undefined,
+				true
+			);
+		});
+
+		it("detects createElement calls created using destructured react cjs require", () => {
```
JoviDeCroock, line 904: Can we add a named `.map` function i.e.

```
function X() {
  return <div>{array.map(function Y() {}})
}
```

robin-drexler: Hey, I’m not sure I understand exactly what you mean here, sorry. 😅 

Could you please elaborate / give a larger example? 

Thank you!

JoviDeCroock: Can you check whether a nested JSX function like i.e. a map inside of the return body leads to any false positives

JoviDeCroock (approved):

PR #693, opened 2025-06-04
support destructured signal access

robin-drexler:
Enable the Babel plugin to register accessing `.value` during destructuring as signals usage. This will end up correctly registering components like the following

```jsx
const X = (props) => {
  const { x: { value: mySignalValue } } = props
}
```

The result is that X will correctly have `useSignals` and react to changing to the x-signal.

### DIFF
--- a/.changeset/sixty-snails-raise.md
+++ b/.changeset/sixty-snails-raise.md
@@ -1,0 +1,5 @@
+---
+"@preact/signals-react-transform": patch
+---
+
+Destructured access to signal values should be registered as usage, before this change Babel would skip a component that would access `.value` on a signal through destructuring
--- a/packages/react-transform/src/index.ts
+++ b/packages/react-transform/src/index.ts
@@ -361,6 +361,19 @@ function isJSXAlternativeCall(
 	return false;
 }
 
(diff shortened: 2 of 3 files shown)


.changeset/sixty-snails-raise.md
```diff
@@ -0,0 +1,5 @@
+---
+"@preact/signals-react-transform": patch
+---
+
+fix destructured signal access not being caught'
```
JoviDeCroock, line 5:
```suggestion
Destructured access to signal values should be registered as usage, before this change Babel would skip a component that would access `.value` on a signal through destructuring
```

packages/react-transform/src/index.ts
```diff
@@ -330,6 +330,25 @@ function isValueMemberExpression(
...
+
+			if (BabelTypes.isIdentifier(key, { name: "value" })) {
+				return true;
+			}
+
+			if (BabelTypes.isObjectPattern(property.value)) {
+				if (hasValuePropertyInPattern(property.value)) {
+					return true;
+				}
+			}
```
JoviDeCroock, line 346: NIT

```suggestion
			if (BabelTypes.isObjectPattern(property.value) && hasValuePropertyInPattern(property.value)) {
			  return true;
			}
```

PR #695, opened 2025-06-07
Refine signal function overloads

wooloo26:


rschristian:
What does this allow you to do that you couldn't otherwise?

wooloo26:
This change itself makes no difference to the user. The issue lies in the inconsistency between direct assignment of the optional parameter `value` and the behavior of the getter, which is a problem in the development implementation. I chose not to modify these aspects because I considered whether they might reflect some special intention by the developers.  

```typescript
declare class Signal<T = any> {
        // optional value
	constructor(value?: T, options?: SignalOptions<T>);
	get value(): T;
}

function Signal(this: Signal, value?: unknown, options?: SignalOptions) {
	this._value = value;
}

Object.defineProperty(Signal.prototype, "value", {
	get(this: Signal) {
		const node = addDependency(this);
		if (node !== undefined) {
			node._version = this._version;
		}
		return this._value;
	}
	...
})
```

PR #696, opened 2025-06-07
Ensure that a state settled hook cant interfere with a signal update

JoviDeCroock:
We should set `component.__f` in Preact again to make sure we hit the branch, this ensures that the new state settling logic as seen in https://github.com/preactjs/preact/pull/4760 can cooperate with signals.

The issue is when we dispatch 3 updates

- SetHookState(new)
- SetSignalState(new)
- SetHookState(old)

This would result in `_afterRender` setting SKIP_CHILDREN for the hook but the signal wants to rightfully update.


JoviDeCroock:
Not needed anymore with https://github.com/preactjs/preact/pull/4829

Issue #395, opened 2023-07-30
Allow effect code to end the effect?


pjeby:
Currently, effects can only be ended from the "outside" - i.e., the caller creating the effect.  But there are times when it's advantageous to allow an effect to cancel itself, since it knows when it is "done".  (In particular, one can have effects decide when to end based on a shared signal, which is the use case I'm working on right now.)

In principle, you can share the dispose callback from the enclosing scope, and perhaps even make a helper function to simplify that process.  But since the effect callback is synchronously invoked, this creates an edge case where you can't dispose the effect before the `effect()` call returns.  Basically, you currently need a helper like this:

```ts
function disposableEffect(fn: (dispose: () => void) => void | (() => void)) {
    let dispose = function () {
        // wait for the effect() call to return
       Promise.resolve().then(() => dispose());
    }
    return dispose = effect(function(){
        return fn.call(this, dispose);
    });
}
```

This seems like it would be greatly simplified by making what is now `_dispose` into a public method, so that effect callbacks can just call `this.dispose()`, or even just return a special value to indicate that they wish to be disposed of.

(Note: while you can certainly stop an effect from having dependencies *eventually*, you can't actually get it to release without first receiving another callback, and if the reason you're trying to end the effect is because there aren't going to *be* any more callbacks, simply dropping all dependencies doesn't work to dispose of the reference cycle between the effect and its final dependencies.)

amb26:
I ran into this myself today - indeed the case where the effect resolve synchronously and hence can't resolve itself in its own scope is annoying to handle. Note that per the "mangled names" registry at https://github.com/preactjs/signals/blob/main/mangle.json#L51 the _dispose method on the effect's prototype is available at the name "d" - and I've seen the preact team condone the use of these names, which have been stable for a good while, in other contexts of "framework building" even if this leaves a somewhat bad taste - e.g. at https://github.com/preactjs/signals/issues/593#issuecomment-2349672856

PR #697, opened 2025-06-11
Publicly allow dispose() on an effect callback

JoviDeCroock:
This enables us to access `this.dispose()` within the callback of an effect. We needed to do some typing shenanigans to allow for an unspecified `this` like we need in our adapter packages so we can override `_start` and others.

Resolves https://github.com/preactjs/signals/issues/395

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -143,6 +143,15 @@ dispose();
 surname.value = "Doe 2";
 ```
 
+Alternatively, you can also declare your effect as a non-arrow function and call `dispose`
+on the `this`.
+
+```js
+effect(function () {
+	this.dispose();
+});
+```
+
 The effect callback may return a cleanup function. The cleanup function gets run once, either when the effect callback is next called _or_ when the effect gets disposed, whichever happens first.
 
 ```js


packages/core/test/signal.test.tsx
```diff
@@ -286,6 +286,48 @@ describe("effect()", () => {
...
+
+		expect(spy).to.be.calledOnce;
+
+		a.value = "aa";
+		expect(spy).to.be.calledOnce;
+
+		a.value = "aaa";
+		expect(spy).to.be.calledOnce;
+	});
+
```
marvinhagemeister, line 330: Nit: Let's add a test where `dispose()` is called from inside the callback and after that the returned dispose function is called. Just to ensure that we're not erroring when it's called twice

packages/core/src/index.ts
```diff
@@ -839,6 +842,9 @@ Effect.prototype._dispose = function () {
 	}
 };
 
+Effect.prototype.dispose = function () {
+	this._dispose();
+};
```
clauderic, line 847: This would be great 🙏 I've got a couple of places in @dnd-kit where I need to set up temporary effects and tear them down from within the effect after a condition is met, for example:

```ts
effect(() => {
  if (manager.dragOperation.status.value === 'dropped') {
    this.dispose();
    performDropAnimation();
  }
});
```

Right now the only way to safely do this is:
```ts
const dispose = effect(() => {
  if (manager.dragOperation.status.value === 'dropped') {
    queueMicrotask(() => dispose());
    performDropAnimation();
  }
});
```

packages/core/test/signal.test.tsx
```diff
@@ -286,6 +286,92 @@ describe("effect()", () => {
...
+	it("should dispose of subscriptions immediately when called twice", () => {
+		const a = signal("a");
+		const b = signal("b");
+		const spy = sinon.spy(() => {
+			a.value + " " + b.value;
+		});
+		const dispose = effect(function () {
+			spy();
+			this.dispose();
```
marvinhagemeister, line 362: Super tiny nit: Maybe also a test where you call `spy()` just after `dispose()`

PR #699, opened 2025-06-11
Add another test

JoviDeCroock:
Follow up to #697

### DIFF
--- a/packages/core/test/signal.test.tsx
+++ b/packages/core/test/signal.test.tsx
@@ -350,6 +350,7 @@ describe("effect()", () => {
 		a.value = "aaa";
 		expect(spy).to.be.calledTwice;
 	});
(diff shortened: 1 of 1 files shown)


packages/core/test/signal.test.tsx
```diff
@@ -350,8 +350,27 @@ describe("effect()", () => {
 		a.value = "aaa";
 		expect(spy).to.be.calledTwice;
 	});
+	it("should dispose of subscriptions immediately and signals are read after disposing", () => {
```
JoviDeCroock, line 353:
```suggestion

	it("should dispose of subscriptions immediately and signals are read after disposing", () => {
```

marvinhagemeister (approved):
Thanks!

PR #706, opened 2025-06-28
feat: dispose effect() with resource management

marvinhagemeister:
Add support for disposing `effect()`'s with the `using` keyword from the stage 3 [explicit resource management](https://github.com/tc39/proposal-explicit-resource-management) proposal.

```js
const count = signal(0);

function doSomething() {
  // The `using` keyword calls dispose at the end of
  // this function scope
  using _ = effect(() => {
    console.log(count.value);
    return () => console.log("disposed");
  });

  console.log("hey");
}

doSomething();
// Logs:
//  0
//  hey
//  disposed
```

Needed to update the babel toolchain to support parsing the `using` keyword in our test setup.


tsconfig.json
```diff
@@ -11,6 +11,7 @@
     "stripInternal": true,
     "noUnusedLocals": true,
     "noEmit": true,
+    "skipLibCheck": true,
```
marvinhagemeister, line 14: Needed to add this because the newer TS version complains about something in sinon types

PR #709, opened 2025-06-28
test: add vitest prep (#707)

JoviDeCroock:
Adds vitest and associated scripts which will not succeed yet, as we have no projects.


43081j:
CI of these will fail right now because it seems vite requires node 20+

we might have to downgrade vite for now, presumably the previous major worked on 18.x

JoviDeCroock:
@43081j currently still trying to figure out why half the tests missbehave on windows

JoviDeCroock:
In that case, good tomerge

PR #724, opened 2025-07-11
feat: add `jsxBind` function to @preact/signals

aleclarson:
I briefly discussed this idea with @rschristian, who said I should try to implement it. Link to the discussion: https://preact.slack.com/archives/C3NMBSJKH/p1752172452437339

The goal is to avoid needing to declare "computed expressions" with `useComputed`, enabling you to write "inline computeds" where they're needed. This is intended for computeds that are only used by one JSX attribute or JSX child.

It's as easy as wrapping a "computed callback" with `jsxBind(…)`, a function exported by the `@preact/signals` package. Only *host elements* support this feature. Currently, there is no warning when trying to use `jsxBind` with a composite element, as this would require iterating the props object. Maybe someone has an idea about how to efficiently check for such misuse.

This implementation supports use in JSX attributes *and* JSX children.

Originally, I named it `bind()`, but decided `jsxBind()` makes its purpose a bit clearer.

### DIFF
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -183,7 +183,7 @@ hook(OptionsTypes.DIFF, (old, vnode) => {
 			if (i === "children") continue;
 
 			let value = props[i];
-			if (value && value.type === Bind) {
+			if (value && value.type === JSXBind) {
 				value = oldSignalProps?.[i] ?? computed(value.cb);
 			}
 			if (value instanceof Signal) {
@@ -472,15 +472,15 @@ export function useSignalEffect(
 	}, []);
 }
 
-function Bind({ cb }: { cb: () => unknown }) {
+function JSXBind({ cb }: { cb: () => unknown }) {
 	return h(SignalValue, {
 		data: useComputed(cb),
 	});
 }
 
-const bindPrototype = Object.getOwnPropertyDescriptors({
+const jsxBindPrototype = Object.getOwnPropertyDescriptors({
 	constructor: undefined,
-	type: Bind,
+	type: JSXBind,
 	get props() {
 		return this;
 	},
@@ -491,8 +491,8 @@ const bindPrototype = Object.getOwnPropertyDescriptors({
  * signals that derive their value from other signals. Like with `useComputed`, any non-signal
  * values used in the callback are captured at the time of binding and won't change after that.
  */
-export function bind<T>(cb: () => T): T {
-	return Object.defineProperties({ cb }, bindPrototype) as any;
+export function jsxBind<T>(cb: () => T): T {
+	return Object.defineProperties({ cb }, jsxBindPrototype) as any;
 }
 
 /**
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -1,4 +1,4 @@
-import { options, Component, isValidElement, Fragment } from "preact";
+import { options, Component, isValidElement, Fragment, h } from "preact";
 import { useRef, useMemo, useEffect } from "preact/hooks";
 import {
 	signal,
@@ -84,7 +84,10 @@ function createUpdater(update: () => void) {
  * A wrapper component that renders a Signal directly as a Text node.
  * @todo: in Preact 11, just decorate Signal with `type:null`
  */
-function SignalValue(this: AugmentedComponent, { data }: { data: Signal }) {
+function SignalValue(
+	this: AugmentedComponent,
+	{ data }: { data: ReadonlySignal }
+) {
 	// hasComputeds.add(this);
 
 	// Store the props.data signal in another signal so that
@@ -172,6 +175,7 @@ Object.defineProperties(Signal.prototype, {
 /** Inject low-level property/attribute bindings for Signals into Preact's diff */
 hook(OptionsTypes.DIFF, (old, vnode) => {
 	if (typeof vnode.type === "string") {
+		const oldSignalProps = vnode.__np;
 		let signalProps: Record<string, any> | undefined;
 
 		let props = vnode.props;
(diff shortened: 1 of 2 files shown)


packages/preact/src/index.ts
```diff
@@ -465,6 +472,29 @@ export function useSignalEffect(
 	}, []);
 }
 
+function JSXBind({ cb }: { cb: () => unknown }) {
```
JoviDeCroock, line 475: What do you think about the following, I am also removing the eager destructuring to save some bytes

```suggestion
type JSXBindable =
  | string
  | number
  | boolean
  | null
  | undefined
  | VNode<any>
  | ComponentChildren;

type JSXBindCallback<T extends JSXBindable = JSXBindable> = () => T;

function JSXBind(props: { cb: JSXBindCallback }) {
```

packages/preact/src/index.ts
```diff
@@ -465,6 +472,29 @@ export function useSignalEffect(
...
+	},
+});
+
+/**
+ * Bind the given callback to a JSX attribute or JSX child. This allows for "inline computed"
+ * signals that derive their value from other signals. Like with `useComputed`, any non-signal
+ * values used in the callback are captured at the time of binding and won't change after that.
+ */
+export function jsxBind<T>(cb: () => T): T {
+	return Object.defineProperties({ cb }, jsxBindPrototype) as any;
```
JoviDeCroock, line 495: Why don't we re-use the existing https://github.com/preactjs/signals/blob/eab99b182985e10a237c53c8a900dbd78e5709d3/packages/preact/src/index.ts#L160-L173 which has the proven way to do this. Currently you _can_ return JSX elements which will exhibit the behaviour we fix in those descriptors

packages/preact/src/index.ts
aleclarson: I managed to remove the `JSXBind` component entirely 👍

packages/preact/src/index.ts
aleclarson: Done 👍

packages/preact/src/index.ts
```diff
@@ -172,13 +178,17 @@ Object.defineProperties(Signal.prototype, {
...
+		const oldSignalProps = vnode.__np;
 		let signalProps: Record<string, any> | undefined;
 
 		let props = vnode.props;
 		for (let i in props) {
 			if (i === "children") continue;
 
 			let value = props[i];
+			if (value && typeof value === "object" && value.__proto__ === jsxBind) {
+				value = oldSignalProps?.[i] ?? computed(value.value);
```
JoviDeCroock, line 190:
```suggestion
				value = oldSignalProps?.[i] || computed(value.value);
```

packages/preact/src/index.ts
```diff
@@ -105,6 +108,9 @@ function SignalValue(this: AugmentedComponent, { data }: { data: Signal }) {
 
 		const wrappedSignal = computed(() => {
 			let s = currentSignal.value.value;
+			if (typeof s === "function") {
```
JoviDeCroock, line 111:
```suggestion
            // This is possibly an inline computed from jsxBind
			if (typeof s === "function") {
```

packages/preact/test/index.test.tsx
```diff
@@ -724,6 +725,105 @@ describe("@preact/signals", () => {
...
+			act(() => {
+				firstName.value = "Jane";
+			});
+
+			// Component should not re-render when only the bound value changes
+			expect(spy).to.have.been.calledOnce;
+			expect(scratch.innerHTML).to.equal("<div>Jane Doe</div>");
+		});
+
+		it("should update bound values without re-rendering the component", async () => {
```
JoviDeCroock, line 786: Can we add a test where we first return JSX and then text or something, similar to the normal SignalValue component

JoviDeCroock (approved):

JoviDeCroock:
This is missing a changeset btw

packages/preact/test/index.test.tsx
aleclarson: Done 👍

packages/preact/src/index.ts
```diff
@@ -172,13 +179,21 @@ Object.defineProperties(Signal.prototype, {
...
 
 		let props = vnode.props;
 		for (let i in props) {
 			if (i === "children") continue;
 
 			let value = props[i];
+			if (
+				value &&
+				typeof value === "object" &&
+				Object.getPrototypeOf(value) === jsxBind
```
JoviDeCroock, line 193: Just using .prototype will be lower bytes and more performant I think, that being said not too fuzzed can look at this later. Will do one final pass over the weekend and get this merged

aleclarson: that wouldn't be equivalent. it's either getPrototypeOf or `.__proto__`

aleclarson: @developit The prototype needs to inherit from `Signal` (see line below jsxBind declaration) for the `SignalValue` component to be used (`jsxBind(…)` expression as JSX child).

personally, I don't see an issue with using `__proto__` as I did originally, but I digress

developit:
Also I believe there is a slight issue with the implementation here (and my suggestions above). In the following example, re-rendering `<Demo />` with a different `class` prop will not re-trigger the computed. Even if the computed is re-triggered (by setting active.value), the callback function is never updated after the first render, so it is closed over the initial props value rather than the current.

```js
let counter = 0;
let someSignal = signal('hello');

function Demo(props) {
  const active = useSignal(false);

  return <div class={jsxBind(() => `${props.class}${active.value ? ' active' : ''}`)} />
}
```

aleclarson:
@developit As I said in the linked Slack thread, I would also prefer not needing the `jsxBind(…)` wrapper call, so happy to see you say that :)

> Also I believe there is a slight issue with the implementation here

That's intentional. It mimics the behavior of `useComputed`

developit:
heh yeah I was just going to point out that this is the same as computed/useComputed.

BTW - does this come up because you're trying to write components that never re-render?  If so, DM me. I ... have been working on a thing. haha.

aleclarson:
@developit sent a Dm on bluesky

PR #727, opened 2025-07-29
Chrome extension stuff

JoviDeCroock:
This is the start of a chrome extension, currently it contains two views. An updates view where we'll stream in updates happening on the page and a second view which contains the signals graph that we can find on the page.

The thing that I'm trying to figure out is a way to display React & Preact components in a framework agnostic way. For instance in the Graph component it would be neat to show which signal is related to which component.

Another thing that needs to happen is adding a Babel plugin to automatically add names to every signal/computed/effect in user-code


extension/manifest.json
```diff
@@ -0,0 +1,39 @@
...
+    {
+      "matches": ["<all_urls>"],
+      "js": ["content.js"],
+      "run_at": "document_start",
+      "all_frames": true
+    }
+  ],
+  "devtools_page": "devtools.html",
+  "background": {
+    "service_worker": "background.js"
```
43081j, line 18: doesn't make much difference but FYI these can be ES modules now:

```ts
  "background": {
    "service_worker": "background.js",
    "type": "module"
  },
```

you don't use anything module-specific but may still be good to know :D

packages/debug/src/index.ts
```diff
@@ -192,9 +220,29 @@ Effect.prototype._callback = function (this: Effect) {
...
+		if (typeof window !== "undefined" && !bridge.shouldThrottleUpdate()) {
+			// Filter updates based on signal names
+			const filteredUpdates = updateInfoList.filter(updateInfo => {
+				const signalName = getSignalName(updateInfo.signal);
+				return bridge.matchesFilter(signalName);
+			});
+
+			if (
+				filteredUpdates.length > 0 &&
+				(window as any).__PREACT_SIGNALS_DEVTOOLS__
```
43081j, line 238: can we strongly type this? might make maintenance easier

```ts
declare global {
  interface Window {
    __PREACT_SIGNALS_DEVTOOLS__?: SignalsDevToolsAPI;
  }
}
```

43081j:
maybe a follow up, but in the UI is there some way we could determine that a change came from the same signal?

similar to how the console in dev tools collapses repeated logged values, it would be useful here.

e.g. if you have a signal that is being changed a _lot_, the panel becomes unusable since it'll just be getting spammed with that one signal. but really its the same signal, so maybe we could collapse it and say something like `a -> b (100x)` (but show the latest change)

PR #750, opened 2025-08-22
feat(extension): display first component parent in render info

43081j:
image

draft for now since we're not sure if we want this yet

something along the lines of this could be useful though, to see what actually rendered. most signals end up in text nodes, so we don't really know the context of the component that actually rendered it. this could help w/ that

### DIFF
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -206,8 +206,15 @@ hook(OptionsTypes.RENDER, (old, vnode) => {
 		typeof window !== "undefined" &&
 		window.__PREACT_SIGNALS_DEVTOOLS__
 	) {
+		let parentComponent: VNode | undefined = vnode.__;
+		while (parentComponent && typeof parentComponent.type === "string") {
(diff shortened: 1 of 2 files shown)


packages/preact/src/index.ts
```diff
@@ -206,8 +206,15 @@ hook(OptionsTypes.RENDER, (old, vnode) => {
 		typeof window !== "undefined" &&
 		window.__PREACT_SIGNALS_DEVTOOLS__
 	) {
+		let parentComponent: VNode | undefined = vnode.__;
+		while (parentComponent && typeof parentComponent.type === "string") {
+			parentComponent = parentComponent.__;
+		}
 		window.__PREACT_SIGNALS_DEVTOOLS__.enterComponent(
-			vnode.type.displayName || vnode.type.name || "Unknown"
+			(parentComponent && typeof parentComponent.type !== "string"
```
43081j, line 214: this code is a bit messy right now. if we actually want this feature ill try tidy it up

PR #752, opened 2025-08-28
feat(extension): support grouped updates

### LINKED ISSUE #736: Repeated updates should collapse
When an update occurs many times in succession we should collapse it, currently it will be

```
signal number
  1 --> 2
  computed doubled
    2 --> 4
signal number
  2 --> 3
  computed doubled
    4 --> 6
```

We'd just collapse this to

```
signal number (2x)
  1 --> 3
  computed doubled
    2 --> 6
```

43081j:
This makes the existing "group updates" setting work.

When enabled, repeated updates of the same tree will be grouped into one node with a count.

The updates are not lost, in that the setting can be disable to re-expand the nodes.

Fixes #736 

---

https://github.com/user-attachments/assets/4b585ea2-f13a-4acd-8b76-9b553eb92f58

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,7 +7,7 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 
-[{*.json,.*rc,*.yml}]
+[{*.json,.*rc,*.yml,*.css}]
 indent_style = space
 indent_size = 2
 insert_final_newline = false
--- a/extension/src/components/UpdateTreeNode.tsx
+++ b/extension/src/components/UpdateTreeNode.tsx
@@ -4,9 +4,10 @@ import { UpdateItem } from "./UpdateItem";
 
 interface UpdateTreeNodeProps {
 	node: UpdateTreeNode;
+	count?: number;
 }
 
-export function UpdateTreeNodeComponent({ node }: UpdateTreeNodeProps) {
+export function UpdateTreeNodeComponent({ node, count }: UpdateTreeNodeProps) {
 	const isCollapsed = useSignal(false);
 
 	const toggleCollapse = () => {
@@ -29,7 +30,7 @@ export function UpdateTreeNodeComponent({ node }: UpdateTreeNodeProps) {
 				)}
 				{!hasChildren && <div className="collapse-spacer" />}
 				<div className="update-content">
-					<UpdateItem update={node.update} />
+					<UpdateItem update={node.update} count={count} />
 				</div>
 			</div>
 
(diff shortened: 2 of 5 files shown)


extension/src/components/UpdateTreeNode.tsx
```diff
@@ -4,9 +4,10 @@ import { UpdateItem } from "./UpdateItem";
 
 interface UpdateTreeNodeProps {
 	node: UpdateTreeNode;
+	count?: number;
 }
 
-export function UpdateTreeNodeComponent({ node }: UpdateTreeNodeProps) {
+export function UpdateTreeNodeComponent({ node, count }: UpdateTreeNodeProps) {
```
43081j, line 10: i really don't like this, so any advice would be appreciated

we want to tell the `UpdateItem` the count, but the `UpdateTreeNode` doesn't really care

if this is the right way of doing that, fair enough

.editorconfig
```diff
@@ -7,7 +7,7 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 
-[{*.json,.*rc,*.yml}]
+[{*.json,.*rc,*.yml,*.css}]
```
43081j, line 10: the CSS files are all formatted with 2 spaces rather than tabs. so i added this to avoid editors adding tabs each time (the pre commit catches it but its still a pain)

extension/src/components/UpdatesContainer.tsx
```diff
@@ -1,11 +1,48 @@
...
+
+	return { tree, counts };
+};
 
 export function UpdatesContainer() {
 	const updatesListRef = useRef<HTMLDivElement>(null);
 	const updateTree = updatesStore.updateTree.value;
+	let collapsedTree = updateTree;
+	let counts: WeakMap<UpdateTreeNode, number> | undefined;
+	if (settingsStore.settings.grouped) {
```
JoviDeCroock, line 43: Let's move this to the `updatesStore`, it could be a computed that only fires when this setting is on. We could also when we collapse change the structure of the tree a bit where `count` is a property

43081j:
if we want to do that, i wonder if we should introduce a `RenderedSignalUpdate` type which would also solve my `count` issue (since the `RenderedSignalUpdate` would have a count, removing the need for the `WeakMap` and `count` props etc)

it could also then have a `firstValue` or some such thing

extension/src/components/UpdatesContainer.tsx
JoviDeCroock: We could replace updateTree and give it a type property so you can return two different shapes. Whatever works best for you

43081j: you can see a rough attempt at this in 69f4641

43081j:
> should we do 2 -> 3 -> 4 or 2 -> 4

i did get this working, but without showing the full path i think it gets confusing

image

this is actually telling us the first value was `0`, a bunch of changes happened, then the value was 5, then 6.

the gap from 0 to 5 i think is more confusing than not having this

but if we do `0 -> 1 -> 2 -> 3 -> 4 -> 5 -> 6` - things will get pretty long, quickly

JoviDeCroock (approved):

Issue #753, opened 2025-09-15
@preact/signals useComputed behavior can be unexpected


jviide:
- [x] Check if updating to the latest version resolves the issue

**Environment**

- [ ] I am using `@preact/signals-core`
- [x] I am using `@preact/signals`
- [ ] I am using `@preact/signals-react`
  - [ ] I understand usage changed in v2, and I've followed the [React Integration instructions](https://github.com/preactjs/signals/tree/main/packages/react#react-integration)

**Describe the bug**

The current @preact/signal's `useComputed` implementation may lead to unexpected behavior when the signals in the compute function's captured scope change. This can make reasoning about useComputed updates hard.

**To Reproduce**

The following link demonstrates the issue: https://stackblitz.com/edit/vitejs-vite-2jxwcfy5?file=src%2Fapp.tsx

1. Observe that the shown value is "1".
2. Click on the "update s1" button, observe that the value is now "2".
3. Click on the "depend on s2" button, observe that the value is still "2".
4. Click on the "update s2 button", observe that the value is still "2".
5. Click on the "update s1 button", observe that the value is now "aa".

As a test-ish pseudocode the flow goes something like this:

```js
function Component({ sig }) {
    return useComputed(() => sig.value);
}

const s1 = signal(1);
const s2 = signal("a");

render(<Component sig={s1} />, scratch);
// now scratch.textContent is "1", the computed depends on s1.

s1.value = 2;
rerender();
// now scratch.textContent is "2", the computed depends on s1.

render(<Component sig={s2} />, scratch);
// scratch.textContent is still "2", the computed still depends on s1.

s2.value = "aa";
rerender();
// scratch.textContent is still "2", the computed still depends on s1.

s1.value = 3;
rerender();
// scratch.textContent is now "aa", the computed depends on s2.
```

**Expected behavior**

There are two main ways to make this behavior more consistent.

 1. Make the compute function one of the `useComputed` dependencies, and update the dependency value on each recompute. This allow the computed to capture potential new signal dependencies (and other non-signal values from the surrounding scope, e.g. prop changes). If the compute function changes on rerender and something is depending on the computed returned by `useComputed`, then this causes a recompute down the line.

    ```js
    function Component({ sig }) {
        return useComputed(() => sig.value);
    }

    const s1 = signal(1);
    const s2 = signal("a");

    render(<Component sig={s1} />, scratch);
    // now scratch.textContent is "1", the computed depends on s1.

    render(<Component sig={s2} />, scratch);
[...]

PR #754, opened 2025-09-15
preact/signals: Update useComputed compute function on rerender

jviide:
This pull request implements alternative 1 for @preact/signals's `useComputed` as described in https://github.com/preactjs/signals/issues/753: Update the compute function on rerender. If the compute function changes, and something is depending on the computed returned by `useComputed`, then this causes a recompute down the line.

This is a mutually exclusive alternative for #755. At least one of these PRs should be eventually closed without merging.


rschristian (approved):
Would need a changeset

PR #755, opened 2025-09-15
preact/signals: Make useComputed always use the initial compute function

jviide:
This pull request implements alternative 2 for @preact/signals's `useComputed` as described in https://github.com/preactjs/signals/issues/753: Always use the initial compute function givent to `useComputed` and only recompute whenever one of the signal dependencies change.

This is a mutually exclusive alternative for #754. At least one of these PRs should be eventually closed without merging.


JoviDeCroock:
Closing this as we settled on the other option

Issue #770, opened 2025-10-10
Add export formats


JoviDeCroock:
**Describe the feature you'd love to see**

We want to optimize the context that can be provided from the extension to i.e. an LLM, to do this we want to enable export to mermaid/export to json to the two views so that folks can copy-paste the input.

PR #774, opened 2025-10-16
feat(extension): support exporting graph

43081j:
This allows the user to export the graph as JSON or mermaid syntax.

image

cc @JoviDeCroock 

Resolves https://github.com/preactjs/signals/issues/770

### DIFF
--- a/extension/src/components/Graph.tsx
+++ b/extension/src/components/Graph.tsx
@@ -34,6 +34,7 @@ export function GraphVisualization() {
 	const isPanning = useSignal(false);
 	const startPan = useSignal({ x: 0, y: 0 });
 	const showExportMenu = useSignal(false);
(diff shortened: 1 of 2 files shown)


extension/src/components/Graph.tsx
```diff
@@ -326,14 +404,42 @@ export function GraphVisualization() {
...
+						>
+							↓ Export
+						</button>
+						{showExportMenu.value && (
+							<div className="graph-export-menu">
+								<button
+									className="graph-export-menu-item"
+									onClick={handleExportMermaid}
+								>
+									Mermaid (clipboard)
```
JoviDeCroock: Rather than displaying (clipboard) can we just do a popup that it got copied to their clipboard?

PR #794, opened 2025-11-12
Delay subscribing to a component until its fully mounted

### LINKED ISSUE #768: Signals never unwatched when using hooks and suspenses

**Environment**

- [x] I am using `@preact/signals-core`
- [x] I am using `@preact/signals-react`
  - [x] I understand usage changed in v2, and I've followed the [React Integration instructions](https://github.com/preactjs/signals/tree/main/packages/react#react-integration)

**Describe the bug**
When using signals in a React project alongside suspenses, signals are never unwatched.

Rather than a wordy description, I've written some unit tests. 

**To Reproduce**

Add the unit tests from this gist to `suspense.test.tsx`

https://gist.github.com/CharlieCrisp/222d10fd147596567a73ae546ff441f1

The following fail:
 - "should clean up signals after unmount with multiple suspense boundaries"
 - "should clean up signals after unmount with multiple suspense boundaries and use of try catch"

**Expected behavior**
Tests should pass, signals should be unwatched when components unmount

JoviDeCroock:
Resolves https://github.com/preactjs/signals/issues/768

This effectively addresses the "missing lifecycle" problem of React, when React suspends during mount it will discard any in progress work which means that the store won't be disposed and hence the subscriptions/... won't be discarded. We circumvent this by overriding the Signal `_subscribe` method,  which sets up the subscriptions, to track all subscriptions for a store and only subscribe them when we are sure the component is fully mounted. This prevents us from subscribing on a work-in-progress state which will just be discarded.

Why do I think this is breaking? React already has a rule not to setState during render but we can't be sure everyone expects that, hence we are marking this as breaking. No synchronous work should be able to intervene during component rendering hence this forking of `Signal.prototype._subscribe` is safe as long as we ensure it's restored in the end.

Thanks to the finally cleaning up with store.f() we are sure this restoration happens correctly. In the manual unmamaged one we might have patterns where we don't which I am fine with due to this being a major bump.

TODO:

- [x] Add tests asserting that `Signal.prototype._subscribe` is restored to its initial form

### DIFF
--- a/packages/react/runtime/src/index.ts
+++ b/packages/react/runtime/src/index.ts
@@ -116,7 +116,7 @@ function startComponentEffect(
 	prevStore: EffectStore | undefined,
 	nextStore: EffectStore
 ) {
-	nextStore._sub = Signal.prototype._subscribe;
+	nextStore._sub = prevStore ? prevStore._sub : Signal.prototype._subscribe;
 	Signal.prototype._subscribe = function (this: Signal, node: any) {
 		nextStore._subscribers.push({ signal: this, node });
 	};
--- a/packages/react/runtime/src/index.ts
+++ b/packages/react/runtime/src/index.ts
@@ -59,6 +59,8 @@ interface Effect {
 	_dispose(): void;
 }
 
+type SubscriptionNode = Parameters<Signal["_subscribe"]>[0];
+
 /**
  * Use this flag to represent a bare `useSignals` call that doesn't manually
  * close its effect store and relies on auto-closing when the next useSignals is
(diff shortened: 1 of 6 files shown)


packages/react/runtime/src/index.ts
```diff
@@ -94,17 +94,24 @@ export interface EffectStore {
...
 }
 
 let currentStore: EffectStore | undefined;
 
+const realSubscribe = Signal.prototype._subscribe;
 function startComponentEffect(
 	prevStore: EffectStore | undefined,
 	nextStore: EffectStore
 ) {
+	nextStore._sub = Signal.prototype._subscribe;
```
JoviDeCroock:
```suggestion
	nextStore._sub = prevStore ? prevStore._sub : Signal.prototype._subscribe;
```

packages/react/runtime/test/browser/suspense.test.tsx
```diff
@@ -123,15 +126,19 @@ describe("Suspense", () => {
 			signal1.value++;
 			signal2.value++;
 		});
+		expect(Signal.prototype._subscribe).to.equal(originalSubscribe);
```
andrewiggins, line 129: I appreciate these checks to make sure the original implementation is restored

andrewiggins (approved):
LGTM

marvinhagemeister (approved):

PR #795, opened 2025-11-13
types: Widen util component types to accept all child node content

rschristian:
Ran into this at work, it'd be nice if you could pass in any `ComponentChild`-compatible content for our util components as both their fallbacks & children. At the moment it's a type error:

```jsx
function Comp() {
  return (
    <Show when={count}>
      'hello'
  {/* ^ 'Show' components don't accept text as child elements. Text in JSX has the type 'string', but the expected type of 'children' is 'Element | ((value: boolean) => Element)'.(2747) */}
    </Show>
  )
}
```

cc @bluetidepro

### DIFF
--- a/packages/react/utils/src/index.tsx
+++ b/packages/react/utils/src/index.tsx
@@ -16,11 +16,11 @@ const Item = (props: any) => {
 		: props.children;
 };
 
-export function Show<T = boolean>(props: ShowProps<T>): ReactNode | null {
+export function Show<T = boolean>(props: ShowProps<T>): JSX.Element | null {
 	useSignals();
 	const value =
 		typeof props.when === "function" ? props.when() : props.when.value;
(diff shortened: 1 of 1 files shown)


packages/react/utils/src/index.tsx
```diff
@@ -20,7 +20,7 @@ export function Show<T = boolean>(props: ShowProps<T>): JSX.Element | null {
 	useSignals();
 	const value =
 		typeof props.when === "function" ? props.when() : props.when.value;
-	if (!value) return props.fallback || null;
+	if (!value) return (props.fallback as JSX.Element) || null;
```
rschristian: This seems to be the correct approach for React's types? Weird: https://github.com/microsoft/TypeScript/issues/14729

If there's a better way lmk

PR #805, opened 2025-12-10
Update types path in package.json for debug package

### LINKED ISSUE #804: `@preact/signals-debug`: types don't resolve

**Environment**

- [x] I am using `@preact/signals-core`

**Describe the bug**
The package.json has an export map which references `dist/debug.d.ts`, but that file doesn't exist. It should probably be `dist/signals-debug.d.ts`.

**To Reproduce**

https://www.typescriptlang.org/play/?#code/MYewdgzgLgBA3hAplAIogRgVwOYHkAOUAluBAL4wC8MAhgO41GxEC2+IATlABQBEAAvg6IawKAHoIRbGBoAbCAFoAJhhy8AlAG4AUElRq8hEpG5wdiWejmJlALhhQOmRABod2DiEz5bDpy46ZNo6FgAe7FzwZFpAA

https://arethetypeswrong.github.io/?p=%40preact%2Fsignals-debug%401.0.1

**Expected behavior**
Types should work automatically based on the export map.

JoviDeCroock:
Resolves #804

### DIFF
--- a/.changeset/grumpy-badgers-tell.md
+++ b/.changeset/grumpy-badgers-tell.md
@@ -1,0 +1,5 @@
+---
+"@preact/signals-debug": patch
+---
+
+Correct types path in package.json for debug package


packages/debug/package.json
jendrikw: This line needs to be changed as well

PR #812, opened 2026-01-14
Add `createModel` API for reactive state containers

andrewiggins:
## Summary

This PR adds `createModel` and `action` to `@preact/signals-core`, providing a structured way to build reactive state containers that encapsulate signals, computed values, effects, and actions.

```js
const CounterModel = createModel((initialCount = 0) => {
	const count = signal(initialCount);
	const doubled = computed(() => count.value * 2);

	effect(() => {
		console.log("Count changed:", count.value);
	});

	return {
		count,
		doubled,
		increment() {
			count.value++;
		},
	};
});

const counter = new CounterModel(5);
counter.increment(); // Updates are automatically batched
counter[Symbol.dispose](); // Cleans up all effects
```

Key features:

- Factory functions can accept arguments for initialization
- All methods are automatically wrapped as actions (batched & untracked)
- Effects created during model construction are captured and disposed when the model is disposed via `Symbol.dispose`
- Models compose naturally - effects from nested models are captured by the parent and disposed together when the parent is disposed
- TypeScript validates that models only contain signals, actions, or nested objects with signals/actions

### No classes or reflection

The implementation avoids using ES classes internally. Using a class would require reflecting onto a class's constructor and the current signals implementation avoids reflection and proxies, so this follows a similar design philosophy. A class-based API could be built on top of this primitive, like so (shoutout @developit for this neat little hack):

```ts
class BaseModelImpl implements Disposable {
	[Symbol.dispose](): void {}
}

export const BaseModel = new Proxy(BaseModelImpl, {
	construct(target, args, newTarget) {
		return createModel(() => Reflect.construct(target, args, newTarget));
	},
}) as unknown as typeof BaseModelImpl;
```

### Using `new` to instantiate models

The public types require using `new` to instantiate models. This helps disambiguate the factory function passed into `createModel` from the returned constructor. It's easier to explain that "`createModel` accepts a factory and returns a class" than "`createModel` accepts a factory and returns a factory."

In other words, this:

```ts
const PersonModel = createModel((name: string) => ({ ... }));
const person = new PersonModel("John");
```

is easier to understand than:

```ts
const createPerson = createModel((name: string) => ({ ... }));
const person = createPerson("John");
```

Using `new` also communicates that each call creates a fresh instance with independent state.

Internally, `createModel` returns a plain function that can be called without `new` for simplicity, but the public types enforce `new` for clarity.

### Automatically capture effects implement a dispose function

Effects declared inside a model's factory function are captured by `createModel` in order to automatically implement a dispose function on the model (exposed as `[Symbol.dispose]`). This design avoids models needing to manually wire up effect dispose from nested models to the model interface.

### Factory functions should NOT return `dispose` functions

If a model needs to run custom logic when it is diposed (that may not be related to signals), it should **not** return a `dispose()` or `[Symbol.dispose]`. When composing models, this dispose function isn't guarenteed to get called as parent models would need to know that your model has a dispose and manually wire it up.

Instead for custom cleanup logic, the recommended pattern is to declare an effect with no signal dependencies that returns a cleanup function that runs the desired cleanup logic. (see "Dispose pattern" below).

### Explicit readonly pattern

Declare your model interface explicitly and use `ReadonlySignal` for signals that should only be modified through actions. This ensures only actions can modify signals, giving you better insight and control over state changes:

```ts
import {
	signal,
	computed,
	createModel,
	ReadonlySignal,
} from "@preact/signals-core";

interface Counter {
	count: ReadonlySignal<number>;
	doubled: ReadonlySignal<number>;
	increment(): void;
	decrement(): void;
}

const CounterModel = createModel<Counter>(() => {
	const count = signal(0);
	const doubled = computed(() => count.value * 2);

	return {
		count,
		doubled,
		increment() {
			count.value++;
		},
		decrement() {
			count.value--;
		},
	};
});

const counter = new CounterModel();
counter.increment(); // OK
counter.count.value = 10; // TypeScript error: Cannot assign to 'value' because it is a read-only property
```

### Dispose pattern

Generally, if you delcare an effect that has cleanup logic, that cleanup logic will before each execution of the effect function (aka whenever the signals your effect relies on update).

However, if you have cleanup logic that needs to run on model dispose that doesn't depend on signals, define an effect that uses no signals but returns your cleanup function. This mirrors the `useEffect(() => { return cleanup }, [])` pattern in React:

```ts
const WebSocketModel = createModel((url: string) => {
	const messages = signal<string[]>([]);
	const ws = new WebSocket(url);

	ws.onmessage = e => {
		messages.value = [...messages.value, e.data];
	};

	// This effect runs once and cleanup is called on dispose
	effect(() => {
		return () => {
			ws.close();
		};
	});

	return {
		messages,
		send(message: string) {
			ws.send(message);
		},
	};
});
```

This pattern is recommended for custom dispose behavior because it allows models to compose naturally - nested models will have their effects cleaned up automatically without manually wiring up dispose functions.

## Open Questions

- Should `createModel` be in its own package? It would require accessing Effect internals to observe effect creation.
- Do people like the `#region` markers? They produce helpful headings in VSCode's scrollbar preview. Happy to remove them if people don't like them.

image

## Future work

- Add `useModel` hook to Preact & React adapters
- Extend debug transform to add names to models & actions, and use model name in signals, computeds, effects, and actions declared within the model
- Extend debug tooling to understand models and actions

### DIFF
--- a/packages/core/test/signal.test.tsx
+++ b/packages/core/test/signal.test.tsx
@@ -9,6 +9,7 @@ import {
 	untracked,
 	ReadonlySignal,
 	ModelConstructor,
+	ModelFactory,
 } from "@preact/signals-core";
 
 describe("signal", () => {
@@ -2950,6 +2951,53 @@ describe("createModel", () => {
 			expect(model.count.value).to.equal(10);
 			model.increment();
 			expect(model.count.value).to.equal(15);
+		});
+
+		it("allows building custom createModel functions that enforces specific arguments", () => {
+			// Demonstrate how to build a custom createModel function
+			// that enforces specific arguments to be passed to the
+			// model factory.
+
+			interface UserContext {
+				userId: string;
+			}
+			type UserContextModelFactory<
+				TModel,
+				TFactoryArgs extends any[] = [],
+			> = ModelFactory<TModel, [UserContext, ...TFactoryArgs]>;
+
+			type UserContextModelConstructor<
+				TModel,
+				TFactoryArgs extends any[] = [],
+			> = ModelConstructor<TModel, [UserContext, ...TFactoryArgs]>;
+
+			type CreateUserModel = <TModel, TFactoryArgs extends any[]>(
+				factory: UserContextModelFactory<TModel, TFactoryArgs>
+			) => UserContextModelConstructor<TModel, TFactoryArgs>;
+
+			const createUserContextModel: CreateUserModel = createModel;
+
+			interface UserModel {
+				userId: ReadonlySignal<string>;
+				setUserId(id: string): void;
+			}
+
+			const UserModel: UserContextModelConstructor<UserModel> =
+				createUserContextModel((context: UserContext) => {
+					const userId = signal(context.userId);
+					return {
+						userId,
+						setUserId(id: string) {
+							userId.value = id;
+						},
+					};
+				});
+
+			const userModel = new UserModel({ userId: "user_1" });
+			expect(userModel.userId.value).to.equal("user_1");
+
+			userModel.setUserId("user_2");
+			expect(userModel.userId.value).to.equal("user_2");
 		});
 	});
 });


docs/demos/todo/index.tsx
```diff
@@ -0,0 +1,258 @@
...
+		// Expose the nested business model
+		todosModel,
+		// View-specific state
+		filter,
+		filteredTodos,
+		setFilter,
+	};
+});
+
+function useModel<TModel>(constructModel: () => Model<TModel>): Model<TModel> {
```
andrewiggins: This hook will get added to the Preact & React adapters in a follow up PR

docs/demos/render-flasher.tsx
```diff
@@ -123,8 +126,8 @@ function flash(nodes: (Node | undefined)[], annotation?: string) {
 			// styled.style.margin = "-2px";
 			styled.style.background = color;
 			styled.style.boxShadow = `0 0 3px 3px ${color}`;
-			styled.style.left = rect.left + "px";
-			styled.style.top = rect.top + "px";
+			styled.style.left = rect.left + scrollX + "px";
+			styled.style.top = rect.top + scrollY + "px";
```
andrewiggins: Little fix for the render flashing to take into account a scrolled window. Ran into this with the model todo app.

packages/core/src/index.ts
```diff
@@ -893,11 +909,142 @@ function effect(fn: EffectFn, options?: EffectOptions): () => void {
 	return dispose as any;
 }
 
+//#endregion Effect
+
+//#region Action
```
JoviDeCroock: I love the regions, we should be good to export this from here because it's a tree-shakeable package

packages/core/src/index.ts
```diff
@@ -893,11 +909,142 @@ function effect(fn: EffectFn, options?: EffectOptions): () => void {
...
+
+/** Models should only contain signals, actions, and nested objects containing only signals and actions. */
+type ValidateModel<TModel> = {
+	[Key in keyof TModel]: TModel[Key] extends ReadonlySignal<unknown>
+		? TModel[Key]
+		: TModel[Key] extends (...args: any[]) => any
+			? TModel[Key]
+			: TModel[Key] extends object
+				? ValidateModel<TModel[Key]>
+				: `Property ${Key extends string ? `'${Key}' ` : ""}is not a Signal, Action, or an object that contains only Signals and Actions.`;
```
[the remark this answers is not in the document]
andrewiggins: Hmmmm I'm torn on this. I think I'd prefer we keep `createModel` exporting raw signals. However I've had some LLMs generate models where constants (like an `id`) are exported from the model. So there might be use case to allow primitives here.

But I'd suggest we keep our guidance to be **not** use getters in models. Use those where you want your model to interface with an existing API - aka at a compatibility layer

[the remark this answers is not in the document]
andrewiggins: I wanna spend an extra moment thinking about the exporting constants though. Would be curious to hear others thoughts too

PR #814, opened 2026-01-15
fix(react-transform): Fix JSX detection leaking to sibling functions

andrewiggins:
# Fix: JSX detection leaking outside of component scope

## The Bug

The Babel transform was incorrectly transforming non-component functions that used signals when a sibling component contained JSX. For example, in the following code:

```js
function wrapper() {
	function Component() {
		return <div>Hello</div>;
	}
	const CountModel = createModel(() => ({
		count: signal(0),
		increment() {
			this.count.value++; // Uses .value
		},
	}));
}
```

The `increment` method inside `CountModel` was being incorrectly transformed because the plugin detected it "contained JSX" due to the sibling `Component` function.

### When This Bug Manifests

This bug requires the component to be nested inside another function scope. The parent function scope is where the `containsJSX` flag would be incorrectly stored, allowing sibling functions to find it when searching their scope chain.

For most application code where components are declared at the module's root scope, this bug wouldn't manifest since there's no parent function scope to hold the leaked flag. However, this is a common pattern in **test files** where components and helper functions are declared inside `describe()` or `it()` blocks—which is how I discovered the bug.

### Root Cause

The bug was multi-faceted:

1. **Function search logic was checking `containsJSX` before it was set**: The `isComponentFunction` helper expected `containsJSX` to already be set on the scope, but it was being called during the function path traversal that would set it. So when looking for a component candidate to mark with `containsJSX`, the current function wouldn't be recognized as a component, causing the search to continue upward.

2. **Fallback to highest-level component**: When searching for a component candidate, the setter function would fallback to the highest level component-like function, which in some scenarios (e.g., test suites with `describe()`) isn't a valid component.

3. **Scope data inheritance**: The critical issue was that `containsJSX` and `maybeUsesSignal` were being set on `scope` rather than on the function node itself. Babel's `scope.getData()` searches parent scopes for data. This meant that when we set `containsJSX` on a parent scope, sibling functions in that same scope would also find that data when querying their own scope, making them appear to "contain JSX" when they didn't.

### Core Change: Store data on function nodes, not scopes

Changed from using `scope.setData()`/`scope.getData()` to using `path.setData()`/`path.getData()` directly on the function node paths. This ensures that `containsJSX` and `maybeUsesSignal` flags are isolated to the specific function that actually contains JSX or signal usage, preventing leakage to sibling functions.

### Refactored `getComponentFunctionDeclaration` to `findParentComponentOrHook`

This function is mostly the same but with a couple bug fixes:

- Find and return the parent component or custom hook function path (not scope)
- Check if a function is a component or hook based on its name, rather than relying on `containsJSX` which may not be set yet
- If a matching component or hook function is now found, return `null` instead of the last seen function.

### Moved `isComponentFunction` check inside `shouldTransform`

Since `isComponentFunction` needs to check `containsJSX`, it should only be called after the function body has been fully traversed. Moving it inside `shouldTransform` (which is called on function exit) ensures the data is available.

## Additional Improvements

- **`getFunctionName` now handles default exports**: The function now takes a `filename` parameter and resolves `DefaultExportSymbol` to the filename internally, simplifying call sites.

- **Added `isCustomHookCallback` helper**: New utility function that detects when a function is being passed as a parameter to a custom hook call (e.g., `useCustomHook(() => { ... })`).

- **Added verbose logging**: New `signals:react-transform:verbose` debug namespace for more detailed logging during development and debugging.

- **Test infrastructure improvements**: Added ability to set `DEBUG_TEST_IDS = false` to skip all generated tests when debugging specific test cases.

### DIFF
--- a/.changeset/proud-tomatoes-flow.md
+++ b/.changeset/proud-tomatoes-flow.md
@@ -1,0 +1,22 @@
+---
+"@preact/signals-react-transform": patch
+---
+
+Fix JSX detection leaking to non-component functions in the same scope
+
+Previously, when a component containing JSX was defined inside another function, the JSX detection could incorrectly "leak" to sibling functions or the parent function, causing non-components to be transformed. This was especially problematic in test files where components are defined inside `it()` or `describe()` blocks.
+
+```js
+describe("suite", () => {
+	it("test", () => {
+		// This arrow function was incorrectly transformed because
+		// Counter's JSX detection leaked to sibling functions
+		const CountModel = () => signal.value;
+		function Counter() {
+			return <div>Hello</div>;
+		}
+	});
+});
+```
+
+The transform now correctly scopes JSX and signal usage detection to only the containing component or custom hook function.


andrewiggins:
Hmm all but one test is passing locally... Not sure why there are more failures here.

Issue #769, opened 2025-10-10
RFC: universal debug extension


JoviDeCroock:
**Describe the feature you'd love to see**

Currently the `extension` is tailored towards a browser-extension, however the UI is quite universally web so we could start leveraging this in `iframe`/`inline`/... there are a lot of ways to work with this so that it could be used in blog-posts and so on.

The moving pieces that we'd need

- The browser extension code as it currently exists
- Extract the communication code and publish it as a `browser-extension-adapter` (the chrome plugin uses this)
- Extract the UI and allow it to be initialized with a communication adapter.

PR #815, opened 2026-01-16
Create separate packages for devtooling

JoviDeCroock:
This creates an independent set of packages including an adapter and ui so that we have all the moving parts for our chrome-extension as well as running it standalone in a browser/iframe/...

Closes https://github.com/preactjs/signals/issues/769

<details>
  <summary>Some screenshots</summary>

image

image

</details>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -55,5 +55,4 @@ jobs:
         working-directory: packages/react
         run: |
           pnpm i react@16 react-dom@16 react-router-dom@5
-          pnpm -w test:minify
-          pnpm -w test:prod
+          pnpm -w exec cross-env COVERAGE=true MINIFY=true vitest run packages/react


package.json
```diff
@@ -3,10 +3,12 @@
   "private": true,
   "scripts": {
     "prebuild": "shx rm -rf packages/*/dist/",
-    "build": "run-p build:*",
+    "build": "run-p build:* && pnpm _build:devtools-ui",
```
JoviDeCroock: This is a separate build step because this library bundles in everything else to avoid in-process duplicates when displaying the UI on the same page as the signals usage.

extension/src/panel.tsx
```diff
@@ -1,57 +1,25 @@
...
-		</div>
-	);
-}
+import { mount } from "@preact/signals-devtools-ui";
+import { createBrowserExtensionAdapter } from "@preact/signals-devtools-adapter";
+import "@preact/signals-devtools-ui/styles";
 
 // Initialize the panel when DOM is loaded
-document.addEventListener("DOMContentLoaded", () => {
+document.addEventListener("DOMContentLoaded", async () => {
```
JoviDeCroock: All the code is gone and you can also see how we'd integrate this generally

- Create adapter
- Mount the UI with the adapter

packages/devtools-adapter/src/postmessage-adapter.ts
```diff
@@ -0,0 +1,159 @@
...
+		this.sendMessage({ type: "REQUEST_STATE" });
+	}
+
+	private sendMessage(message: any): void {
+		this.targetWindow.postMessage(message, this.targetOrigin);
+	}
+
+	private handleMessage(event: MessageEvent): void {
+		// Validate origin
+		if (event.origin !== this.sourceOrigin) return;
```
developit: Names are a bit odd here cause `targetWindow` is the "target" for sending messages and this method is for receiving messages, but if we check `event.source` we could probably just completely ignore the origin or make it optional.

```suggestion
		if (event.origin !== this.sourceOrigin) return;
		if (event.source !== this.targetWindow) return;
```

PR #816, opened 2026-01-17
Try an alternative approach to component tracking

JoviDeCroock:
This tries to provide a better way to subscribe and especially unsubscribe components in the devtools.

Before we had a very manual `mount`/`unmount` sequence but this ended up being hard to track. Now instead of doing all of the manual labour we consider the `effect` a special case.


JoviDeCroock:
This approach is currently not working, it was a nice experiment though

PR #865, opened 2026-02-02
Fix scu bail

JoviDeCroock:
Due to us leveraging useState the shouldComponentUpdate override isn't effective


developit:
@JoviDeCroock fix here that would prevent the need to touch Prefresh would just be to use useRef

```js
const ref = useRef();
let s = ref.current;
if (!s) ref.current = s = signal(value);
return s;
```

PR #889, opened 2026-02-22
With our patched rename-properties plugin we surfaced a name collission

JoviDeCroock:
Not sure how this worked before we mapped both _source and _subscribe to S, I chose to re-map _subscribe because the only user of that is the debug package. I am mildly concerned how this hasn't surfaced before.

My thinking is that somehow the terser mangling differentiates in this


JoviDeCroock:
Never mind, I think my testing setup was just acting up

Issue #242, opened 2022-10-15
"No-op" batch triggers effects


felixschorer:
Given a signal `foo` and an effect logging its value.
When I run a batch, which changes the value of `foo` in such a way that it is essentially a no-op,
then the value of `foo` is logged even though the value of `foo` is the same as before the batch.
```ts
const foo = signal(42);

effect(() => console.log(foo.value));

batch(() => {
  foo.value = 0;
  foo.value = 42;
});
```

This isn't the case when the effect depends on a computed signal.
```ts
const foo = signal(42);
const bar = computed(() => foo.value);

effect(() => console.log(bar.value));

batch(() => {
  foo.value = 0;
  foo.value = 42;
});
```

eddyw:
As I understand by working with the internals the last weeks, `batch` doesn't batch updates but rather batches effects, it works sort of like this:
```
batch(() => {
  foo.value = 0;
  // ^ value changed, it increments the signal version number +1
  // ^ foo notifies all targets
  // ^     if target is a computed with subscribers (e.g: an effect),
  // ^          forward notification to subscribers (other effects/computed) (repeat)
  // ^     else if target is an effect
  // ^          effect is batched (added as a head of a LinkedList of effects to run)
  foo.value = 42;
  // ^ repeat same as above, version number increments +1 again
})
// ^ end of batch:
// Walk through each batched effect,
//   Check if needs to re-run by comparing the version number of its sources (signals/computed)
//   While doing this, if source is a 'computed',
//       Computed checks if need to re-run by comparing version number of its sources
//       Re-run computed and increase its version number +1 ... or not
//   Re-run effect .. or not
```

Basically, signals use a 'version' number rather than the actual value to know when a 'computed' or 'effect' needs to re-run. This is done because comparing numbers is cheaper. The "downside" is that if you set the same value as it was before the batch, there is no way to know if the value is the same, again because it compares version numbers which increase of every set and not the actual value.

There is this comment in the code regarding the version:
```
	// The version number of the source that target has last seen. We use version numbers
	// instead of storing the source value, because source values can take arbitrary amount
	// of memory, and computeds could hang on to them forever because they're lazily evaluated.
	// Use the special value -1 to mark potentially unused but recyclable nodes.
```

felixschorer:
I think you're confusing nodes with signals. Nodes only have a version. Signals and computed signals have a version and keep a reference to the current / cached value.

[...]

PR #891, opened 2026-02-23
Propose way to not have useless effect/computed runs

JoviDeCroock:
Resolves https://github.com/preactjs/signals/issues/242

This effectively allows us to not run effects that didn't end up changing in a batch

### DIFF
--- a/.changeset/big-fans-visit.md
+++ b/.changeset/big-fans-visit.md
@@ -1,0 +1,5 @@
+---
+"@preact/signals-core": minor
+---
+
+Prevent batches where a signal goes from A --> B --> A from triggering dependent updates, a computed/effect should not re-run when the dependencies in a batched update amount to an equal value.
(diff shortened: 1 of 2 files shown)


rschristian (commented):
Have we configured different browser targets for CJS & ESM output? The size increase on `packages/core/dist/signals-core.js` looks like it's polyfilling something maybe? The ESM size diff, on the other hand, looks more like I'd expect.

At work so I can't actually install & build the repo

rschristian:
My comment revision history should show that I originally guessed it'd be `Map`, then backed off as I couldn't quite remember how old it was lol

Strange though, Microbundle shouldn't be treating the CJS & ESM builds any differently in terms of target, but I guess that's just another reason in the long list of reasons why it'd be great to drop Microbundle.

rschristian (approved):

PR #904, opened 2026-03-15
feat(react-transform-rolldown): add native rolldown plugin

JoviDeCroock:
migrated from https://github.com/rolldown/plugins/pull/21

## Summary
- add a new `@preact/signals-react-transform-rolldown` package with version `0.0.0`
- port the native Rolldown React Signals transform and its transform/runtime coverage into the Signals monorepo
- add workspace wiring and a minor changeset so the package can be built, tested, and released from this repo

## Testing
- `pnpm --filter @preact/signals-react-transform-rolldown test`
- `pnpm build:react-transform-rolldown`

### DIFF
--- a/.changeset/small-peaches-wave.md
+++ b/.changeset/small-peaches-wave.md
@@ -1,0 +1,5 @@
+---
+"@preact/signals-react-transform-rolldown": minor
+---
+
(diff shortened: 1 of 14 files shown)


marvinhagemeister (approved):
LGTM

marvinhagemeister (approved):

PR #909, opened 2026-03-29
Extract key from child-element

JoviDeCroock:
Follow up to #907 CC @jbalsas


jbalsas (approved):
Makes sense to me! Is there no way we could do this without having to wrap the passed in children in `<Item>` but still keep the scoped reactivity to the items? 🤔

marvinhagemeister (approved):

JoviDeCroock:
I realised this is a bit dangerous to running props.children() eagerly which would mean accessed signals within a child would re-render the whole For and hit the cache issue more

PR #918, opened 2026-05-06
Fix live-signal updating other components from render

JoviDeCroock:
The famous `can't update component x from the render of component y` is back, and it's because we are indeed updating a signal during render :D


rschristian (approved):

marvinhagemeister (approved):
LGTM

Issue #614, opened 2024-10-23
Re-assigning NaN is seen as a value change


Legioth:
- [x] Check if updating to the latest version resolves the issue

**Environment**
- [x] I am using `@preact/signals-core`
- [ ] I am using `@preact/signals`
- [ ] I am using `@preact/signals-react`
  - [ ] I understand usage changed in v2, and I've followed the [React Integration instructions](https://github.com/preactjs/signals/tree/main/packages/react#react-integration)

**Describe the bug**
In JavaScript, `NaN === NaN` evaluates to `false`. A consequence of this is that a signal with the value `NaN` will trigger an update if the value is again assigned as `NaN`. In the worst case, this might lead to an infinite loop with a two-way UI binding if the receiving end is also treating `NaN` in the same way.

**To Reproduce**

```js
const x = signal(0);
effect(() => console.log(x.value));
x.value = NaN;
x.value = x.value;
```

This logs `NaN` twice even though it's expected that the last row shouldn't lead to running the effect again.

**Expected behavior**
Expected that the dirty checking logic has a special case for re-assigning `NaN`

Legioth:
Just for context, I discovered through through doing `parseInt(fieldValue)` for an empty string and then wasted some debugging time looking at other things before realizing that `NaN != NaN`. More graceful handling in the signal would have helped me avoid an infinite loop which would in turn have made it easier for me to realize that I had an unexpected "number" in the system.

rschristian:
You should really never call `parseInt` without having a fallback though is my point.

I believe the intention here is to still ship ES5, so `Object.is()` is likely out, and anything else would probably hurt perf slightly. We also don't support it in Preact for that reason.

---

Again, not saying we won't, but I at least would prefer not to. It's a bad pattern to pass `NaN` around, IMO.

dmail:
I agree it's more the job of a linter and it's a bad pattern to have NaN.
But this is an easy mistake to make. I think if preact signals can help human devs to catch errors and gain time it should.

PR #943, opened 2026-06-13
fix(core): treat reverting a batched signal to NaN as a no-op

chatman-media:
## Summary

The no-op-batch optimization (added in c0caa20, "Propose way to not have useless effect/computed runs") is supposed to ensure that writing a signal and then reverting it to its original value **within the same `batch()`** does not re-run subscribers. This works for normal values but silently breaks for `NaN`.

`reconcileBatchSnapshots()` decides whether to roll back a signal's version by comparing the final value to the pre-batch snapshot with `===`:

```js
if (snapshots._source._value === snapshots._value) {
    snapshots._source._version = snapshots._version;
}
```

Since `NaN === NaN` is `false`, a signal that started as `NaN` and is reverted back to `NaN` inside a batch never has its version rolled back, so subscribed effects re-run even though nothing changed.

## Reproduction

```js
const foo = signal(NaN);
effect(() => { foo.value; }); // runs once

batch(() => {
  foo.value = 0;
  foo.value = NaN; // revert to original
});
// effect re-runs — but the equivalent non-NaN case does not:
//   batch(() => { foo.value = 0; foo.value = 42; })  // already suppressed
```

This is the batch-level counterpart of the asymmetry described in #614. It is scoped strictly to the no-op-batch optimization and does **not** change the plain `.value` setter semantics.

## Fix

Add a self-comparison so a value that started and ended as `NaN` is treated as unchanged. `x !== x` is `true` only for `NaN`, so this avoids `Object.is` and keeps the ES5 output:

```js
if (value === snapshot || (value !== value && snapshot !== snapshot)) {
    snapshots._source._version = snapshots._version;
}
```

A real change *to* `NaN` (e.g. `1` → `NaN` in a batch) still correctly re-runs effects — covered by a test.

## Tests

- `should not rerun an effect for a no-op batch assignment that reverts to NaN` — fails before the fix, passes after.
- `should rerun an effect for a batch assignment that changes the value to NaN` — guards the real-change direction.

Both live next to the existing `should not rerun an effect for a no-op batch assignment` test. Full `packages/core` suite is green (158 passing, 2 pre-existing skips).

Related: #614

### DIFF
--- a/.changeset/fix-batch-noop-nan-revert.md
+++ b/.changeset/fix-batch-noop-nan-revert.md
@@ -1,0 +1,5 @@
+---
+"@preact/signals-core": patch
+---
+
+Fix a no-op `batch` assignment that reverts a signal back to `NaN` incorrectly re-running effects. The batch snapshot reconciliation compared the final value to the pre-batch value with `===`, which is always `false` for `NaN`, so the version was never rolled back. Reverting a `NaN` value to `NaN` within a batch is now correctly treated as unchanged.
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -167,7 +167,13 @@ function reconcileBatchSnapshots() {
 	batchSnapshots = undefined;
 
 	while (snapshots !== undefined) {
-		if (snapshots._source._value === snapshots._value) {
+		const value = snapshots._source._value;
+		const snapshot = snapshots._value;
+		// Roll back the version if the value ended up unchanged. The extra
+		// self-comparison treats a value that started and ended as NaN as
+		// unchanged, since `NaN === NaN` is `false` (avoids `Object.is` to
+		// keep ES5 output).
+		if (value === snapshot || (value !== value && snapshot !== snapshot)) {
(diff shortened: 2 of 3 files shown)


rschristian:
Profile reeks of being an LLM which, undisclosed, is against our [Code of Conduct](https://github.com/preactjs/.github/blob/master/CODE_OF_CONDUCT.md):

> We recognize that LLM's have become an integral tool for many users, however when using LLMs to help write issues or pull requests, disclose this. Transparency preserves the collaborative integrity that makes open-source communities work. It's the difference between "I used AI to help structure this bug report" (honest, helpful) and silently passing off generated content as your own analysis (deceptive, counterproductive). We as maintainers want to engage with humans rather than LLM-generated messages, put time in your issue report and that will be reflected back.

Additionally, it makes little sense to special-case batches, every value should match in behavior.

---

We'll push out a new version at some point using `Object.is`, likely alongside or after Preact 11. I don't think we should change this w/ that being on the horizon.

PR #945, opened 2026-06-16
fix(core): prevent model effect capture in untracked

andrewiggins:
## Summary

- Prevent `untracked()` from propagating ambient `createModel()` effect capture into its callback
- Ensure `action()` callbacks get the same behavior because `action()` delegates to `untracked()`
- Add coverage for `untracked()`, `action()`, and nested model ownership boundaries

Co-authored by: Pi coding agent & GPT 5.5

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-npm run nano-staged
+pnpm run nano-staged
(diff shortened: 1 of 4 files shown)


.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-npm run nano-staged
+pnpm run nano-staged
```
JoviDeCroock, line 4: Looks pretty driveby

JoviDeCroock (approved):

marvinhagemeister (approved):

.husky/pre-commit
andrewiggins: Heh fair. I intentionally added this cuz we use pnpm for all of our commands elsewhere in the repo, so figured I clean up this one last usage of npm.

PR #950, opened 2026-07-06
Allow For to accept readonly arrays

joeyTedeschi:
## Summary
- Expand `<For>` accepted list types to `ReadonlyArray<T>` for direct arrays, signals, readonly signals, and function-returned list sources
- Keep Preact and React utility implementations in sync
- Add browser utility tests for direct readonly arrays, `signal<readonly T[]>`, and computed readonly signals
- Add a patch changeset for `@preact/signals` and `@preact/signals-react`

Without this change, attempting to use a `readonly` array with `For` results in a `TS 2322` (due to assigning `readonly T[]` to `T[]`).

PR was created in collaboration with AI (Codex GPT 5.5), which was used to generate the `.changeset/readonly-for-inputs.md` and unit tests. PR was opened in draft, manually audited, and then opened for wider review

### DIFF
--- a/packages/preact/utils/src/index.tsx
+++ b/packages/preact/utils/src/index.tsx
@@ -28,11 +28,13 @@ export function Show<T = boolean>(
 
 Show.displayName = "Show";
 
+type ForEach<T> =
+	| ReadonlyArray<T>
+	| Signal<ReadonlyArray<T>>
+	| ReadonlySignal<ReadonlyArray<T>>;
+
 interface ForProps<T> {
-	each:
-		| Signal<Array<T>>
-		| ReadonlySignal<Array<T>>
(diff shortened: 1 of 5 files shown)


packages/preact/utils/src/index.tsx
```diff
@@ -28,11 +28,13 @@ export function Show<T = boolean>(
 
 Show.displayName = "Show";
 
+type ForEach<T> =
+	| ReadonlyArray<T>
```
joeyTedeschi, line 32: This means that we'd now accept normal arrays as a valid type as well. The `const list` + `listValue` logic already supported, but is there a concern with supporting it explicitly?

JoviDeCroock (approved):

PR #959, opened 2026-07-22
Name internal Preact signals

JoviDeCroock:
Internal signals created by the Preact adapter and `useLiveSignal` appear under minified names in debugging tools.

Assign stable `ReactiveDom` and `useLiveSignal` names so they are identifiable.

### DIFF
--- a/packages/preact/src/index.ts
+++ b/packages/preact/src/index.ts
@@ -108,6 +108,7 @@ function SignalValue(this: AugmentedComponent, { data }: { data: Signal }) {
 	// Store the props.data signal in another signal so that
 	// passing a new signal reference re-runs the text computed:
 	const currentSignal = useSignal(data);
+	currentSignal.name = "ReactiveDom";
 	currentSignal.value = data;
 
 	const [isText, s] = useMemo(() => {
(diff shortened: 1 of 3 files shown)


packages/preact/src/index.ts
```diff
@@ -107,7 +107,7 @@ function SignalValue(this: AugmentedComponent, { data }: { data: Signal }) {
 
 	// Store the props.data signal in another signal so that
 	// passing a new signal reference re-runs the text computed:
-	const currentSignal = useSignal(data);
+	const currentSignal = useSignal(data, { name: "ReactiveDom" });
```
marvinhagemeister: Nit: Since `name` is a public property we could set it directly and avoid the object. That said not sure if it matters and engines optimize that away anyways.

marvinhagemeister (approved):

packages/preact/src/index.ts
JoviDeCroock: Good point — addressed in the rebase. Now setting `currentSignal.name = "ReactiveDom"` and `s.name = "useLiveSignal"` directly instead of allocating the options object.

PR #966, opened 2026-07-31
Export ModelFactory from `@preact/signals`

BPScott:
Reexport ModelFactory (along with the already exported Model and ModelConstructor) from `@preact/signals`


JoviDeCroock (approved):
Just needs a changeset

BPScott:
Thanks for adding the changeset Jovi! That was a fast-and-dirty edit-in-github's-ui

PR #975, opened 2026-08-14
Unwrap signals returned from Show's when callback

marvinhagemeister:
## Problem

When `Show`'s `when` prop is a callback like `() => someSignal`, the returned Signal object was used directly in the truthiness check. Since Signal objects are always truthy, the fallback branch was never rendered — even when the signal's value was `false`, `0`, ``, or `null`.

This is inconsistent with the non-callback form `when={someSignal}`, which already reads `.value`.

## Fix

Unwrap the return value when it's a Signal, reading `.value` before the truthiness check:

```tsx
const raw = typeof props.when === "function" ? props.when() : props.when.value;
const value = raw instanceof Signal ? raw.value : raw;
```

## Why this matters

SFC compilers (like `@preact-labs/sfc-codegen`) emit `{#if expr}` blocks as `<Show when={() => expr}>`. When `expr` is a signal, the callback returns the Signal object. Without unwrapping, `{#if falsySignal}` would always render the consequent branch instead of the fallback.

### DIFF
--- a/packages/preact/utils/src/index.tsx
+++ b/packages/preact/utils/src/index.tsx
@@ -20,8 +20,12 @@ Item.displayName = "Item";
 export function Show<T = boolean>(
 	props: ShowProps<T>
 ): ComponentChildren | null {
-	const value =
+	const raw =
 		typeof props.when === "function" ? props.when() : props.when.value;
+	// Unwrap a signal returned by a `when` callback so `when={() => someSignal}`
+	// checks the signal's value, not the signal object (which is always truthy).
+	const value =
+		raw instanceof Signal ? (raw as unknown as Signal<T>).value : raw;
 	if (!value) {
 		const fallback = props.fallback;
 		return typeof fallback === "function" ? fallback() : fallback || null;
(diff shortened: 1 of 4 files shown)


packages/preact/utils/src/index.tsx
```diff
@@ -20,8 +20,12 @@ Item.displayName = "Item";
 export function Show<T = boolean>(
 	props: ShowProps<T>
 ): ComponentChildren | null {
-	const value =
+	const raw =
 		typeof props.when === "function" ? props.when() : props.when.value;
+	// Unwrap a signal returned by a `when` callback so `when={() => someSignal}`
+	// checks the signal's value, not the signal object (which is always truthy).
+	const value =
```
JoviDeCroock, line 27: I am not a big fan of this, this feels like just adding bytes. The callback was intended to be a shortcut to a computed

marvinhagemeister: fair point, I'll move that to the SFC compiler

PR #978, opened 2026-08-29
Reduce Preact prop updater allocations

JoviDeCroock:
Each signal-bound DOM property currently allocates a wrapper object, updater closure, effect callback closure, and bound disposer. Make the Effect itself the property updater and share its callback and update methods across bindings, while retaining the wrapper Signal that preserves rebinding semantics.

Trade-off: signals.mjs grows from 1732 B to 1821 B gzip (+89 B). Each live binding is 72 B smaller, so the shipped cost breaks even after a few bound properties.

<details>
<summary>Signal-bound DOM properties — 52.5% faster creation, 72 B smaller, 11.1% faster updates</summary>

```jsx
import { h, render } from "preact";
import { signal } from "@preact/signals";

const count = 10_000;
const values = Array.from({ length: count }, (_, i) => signal(`before-${i}`));
const root = document.createElement("div");
document.body.append(root);

function App() {
  return h(
    "main",
    null,
    values.map((value, i) => h("span", { title: value }, i))
  );
}
render(h(App), root);

async function write(value) {
  const start = performance.now();
  for (let i = 0; i < count; i++) values[i].value = `${value}-${i}`;
  await Promise.resolve();
  await Promise.resolve();
  await Promise.resolve();
  return performance.now() - start;
}

for (let i = 0; i < 5; i++) await write(`warm-${i}`);
const samples = [];
for (let i = 0; i < 15; i++) samples.push(await write(`sample-${i}`));
console.log(Math.min(...samples));
```
</details>


JoviDeCroock:
Byte size impact is a bit too large

